PR #10, opened 2022-03-15
'Solution'

serhii56465:


services/movie_session.py
```diff
@@ -0,0 +1,76 @@
...
+def get_movie_session_by_id(movie_session_id: int):
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time=None,
+        movie_id: int = None,
+        cinema_hall_id: int = None):
+    queryset = MovieSession.objects.all()
```
roland813, line 31: You can filter queryset at once and don't do that in every loop

services/movie_session.py
```diff
@@ -0,0 +1,76 @@
...
+
+    if not show_time and movie_id and not cinema_hall_id:
+        queryset.filter(id=session_id).update(
+            movie=movie_id
+        )
+
+    if not show_time and not movie_id and cinema_hall_id:
+        queryset.filter(id=session_id).update(
+            cinema_hall=cinema_hall_id
+        )
```
roland813, line 71: You can check separately if any of arguments get. And separatly update, less loops

roland813 (changes requested):

MasterpieceElbow (approved):

PR #20, opened 2022-04-07
Solution

IgorOlifirenko:


services/movie_session.py
```diff
@@ -0,0 +1,44 @@
...
+    if movie_id:
+        movie_session.update(movie_id=movie_id)
+
+    if show_time:
+        movie_session.update(show_time=show_time)
+
+    if cinema_hall_id:
+        movie_session.update(cinema_hall_id=cinema_hall_id)
+
+    return movie_session
```
vldtsw, line 40: No need to return anything here

services/movie_session.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def delete_movie_session_by_id(session_id: int):
+    return MovieSession.objects.filter(id=session_id).delete()
```
vldtsw, line 44: And here

vldtsw (changes requested):

IvanRamyk (approved):
Well done!

PR #26, opened 2022-06-18
Solution

YuriiKindrat:


services/movie.py
```diff
@@ -0,0 +1,33 @@
...
+                 movie_description: str,
+                 genres_ids: list[int] = None,
+                 actors_ids: list[int] = None):
+
+    new_movie = Movie.objects.create(title=movie_title,
+                                     description=movie_description)
+
+    if genres_ids:
+        for id_ in genres_ids:
+            new_movie.genres.add(Genre.objects.get(id=id_))
```
rotsen18, line 29: You may just add genres by ids. You don't need for loop with querues to db for that.

services/movie.py
```diff
@@ -0,0 +1,33 @@
...
+
+    if actors_ids:
+        for id_ in actors_ids:
+            new_movie.actors.add(Actor.objects.get(id=id_))
```
rotsen18, line 33: same

services/movie_session.py
```diff
@@ -0,0 +1,48 @@
...
+
+def create_movie_session(movie_show_time,
+                         movie_id: int,
+                         cinema_hall_id: int):
+
+    movie = Movie.objects.get(id=movie_id)
+    cinema_hall = CinemaHall.objects.get(id=cinema_hall_id)
+    MovieSession.objects.create(show_time=movie_show_time,
+                                cinema_hall=cinema_hall,
+                                movie=movie)
```
rotsen18, line 12: Here you can create MovieSession without creating movie, cinema_hall objects. Generaly it works, but you have not necessary queries to db

services/movie_session.py
```diff
@@ -0,0 +1,48 @@
...
+                         cinema_hall_id=None):
+
+    session_for_update = MovieSession.objects.filter(id=session_id)
+
+    if show_time:
+        session_for_update.update(show_time=show_time)
+
+    if cinema_hall_id:
+        new_cinema_hall = CinemaHall.objects.get(id=cinema_hall_id)
+        session_for_update.update(cinema_hall=new_cinema_hall)
```
rotsen18, line 40: same

services/movie_session.py
```diff
@@ -0,0 +1,48 @@
...
+
+    if movie_id:
+        new_movie = Movie.objects.get(id=movie_id)
+        session_for_update.update(movie=new_movie)
```
rotsen18, line 44: same

rotsen18 (approved):
In general, everything works, but if you want you can try to reduce the number of queries to the database

services/movie.py
MasterpieceElbow, line 29: You can pass a few ids into the `.add()`: `new_movie.genres.add(1, 2, 3)` so you don't need to use loop and queries to get genres.

services/movie.py
MasterpieceElbow, line 33: Same here

services/movie_session.py
MasterpieceElbow, line 12: You don't need to get movie and cinema_hall objects. You can only pass their ids to `cinema_hall_id` and `movie_id` arguments.

services/movie_session.py
MasterpieceElbow, line 40: The same here

services/movie_session.py
MasterpieceElbow, line 44: And here

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #28, opened 2022-06-20
'movie_moviesession_cinemahall'

CheshkoSergey:


services/movie_session.py
```diff
@@ -0,0 +1,50 @@
+from db.models import MovieSession
+
+
+def create_movie_session(
+        movie_show_time,
```
katevasilenko, line 5: _movie_show_time: str_ I guess

katevasilenko (approved):

PR #31, opened 2022-06-20
solution

MarkoKhodan:


db/models.py
```diff
@@ -14,3 +14,39 @@ class Actor(models.Model):
...
+    actors = models.ManyToManyField(Actor, related_name="movies")
+    genres = models.ManyToManyField(Genre, related_name="movies")
+
+    def __str__(self):
+        return self.title
+
+
+class CinemaHall(models.Model):
+    name = models.CharField(max_length=255)
+    rows = models.IntegerField(default=1)
```
IvanRamyk, line 31: Not sure that 1 as a default value is a good idea for this model. 
If `rows` or `seats_in_row` are not provided it's better to get an error than unexpected values in the corresponding fields 
I believe, it's better to remove the default value.

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

IvanRamyk (approved):
Well done!

PR #40, opened 2022-06-21
'Solution

SergeyDev1996:


db/models.py
```diff
@@ -2,15 +2,51 @@
...
+
+    def __str__(self):
+        return self.title
+
+
+class CinemaHall(models.Model):
+
+    name = models.CharField(max_length=255)
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField(default=1)
```
Yurnerroo, line 34: default value is not necessary

db/services/movie.py
```diff
@@ -0,0 +1,33 @@
+from db.models import Movie
+
+
+def get_movies(genres_ids, actors_ids):
+    queryset = Movie.objects.all()
+    if genres_ids is None and actors_ids is None:
+        return queryset
+    if genres_ids is not None and actors_ids is not None:
+        queryset = queryset.filter(genres__id__in=genres_ids)
+        queryset = queryset.filter(actors__id__in=actors_ids)
```
Yurnerroo, line 10: you can write few operands in one line, for example: filter(genres__id__in=genres_ids, actors__id__in=actors_ids)

db/services/movies_session.py
```diff
@@ -0,0 +1,36 @@
...
+def get_movies_sessions(session_date=None):
+    sessions = MovieSession.objects.all()
+    if session_date:
+        sessions.filter(show_time=session_date)
+    return sessions
+
+
+def get_movie_session_by_id(movie_session_id):
+    movie = MovieSession.objects.filter(id=movie_session_id).all()
+    return movie
```
Yurnerroo, line 19: you can just use get() instead of 2-step operation with filter+all

services/movie.py
```diff
@@ -0,0 +1,33 @@
+from db.models import Movie
+
+
+def get_movies(genres_ids=None, actors_ids=None):
+    queryset = Movie.objects.all()
+    if genres_ids is None and actors_ids is None:
+        return queryset
+    if genres_ids is not None and actors_ids is not None:
+        queryset = queryset.filter(genres__id__in=genres_ids)
+        queryset = queryset.filter(actors__id__in=actors_ids)
```
Yurnerroo, line 10: again, there is an opportunity to use 1 filter with 2 checks

services/movie_session.py
```diff
@@ -0,0 +1,36 @@
+from db.models import MovieSession
+
+
+def create_movie_session(movie_show_time, movie_id, cinema_hall_id):
+    create = MovieSession.objects.create(show_time=movie_show_time,
+                                         movie_id=movie_id,
+                                         cinema_hall_id=cinema_hall_id)
+    return create
```
Yurnerroo, line 8: return is not necessary in such functions

Yurnerroo (approved):

db/services/movie.py
MasterpieceElbow, line 10: You can make it as a single query.

services/movie.py
```diff
@@ -0,0 +1,33 @@
...
+    queryset = Movie.objects.all()
+    if genres_ids is None and actors_ids is None:
+        return queryset
+    if genres_ids is not None and actors_ids is not None:
+        queryset = queryset.filter(genres__id__in=genres_ids,
+                                   actors__id__in=actors_ids)
+    if genres_ids is not None and actors_ids is None:
+        queryset = queryset.filter(genres__id__in=genres_ids)
+    if actors_ids is not None and genres_ids is None:
+        queryset = queryset.filter(actors__id__in=actors_ids)
```
MasterpieceElbow, line 14: It is a bad approach. You can simply use `if genres_ids` and `if actors_ids` to filter the previous queryset.

MasterpieceElbow (changes requested):

TPolina (changes requested):
Why do you have duplicate files?

TPolina (approved):

PR #41, opened 2022-06-21
'Solution'

Yurnerroo:


services/movie.py
```diff
@@ -0,0 +1,29 @@
+from db.models import Movie
+
+
+def get_movies(genres_ids=None, actors_ids=None):
+    if genres_ids is None and actors_ids is None:
+        return Movie.objects.all()
```
SergeyDev1996, line 6: as per video, we need to use some variable(queryset), and later we need to filter the queryset. However, I don't know how it will influence the code.

services/movie.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def create_movie(movie_title, movie_description,
+                 genres_ids=None, actors_ids=None):
+    film = Movie.objects.create(
+        title=movie_title,
+        description=movie_description
+    )
+    if genres_ids is not None:
+        film.genres.add(*genres_ids)
```
SergeyDev1996, line 27: as per video, we were advice to use the "set", not add. But if it works, not sure if there is a reson to change it.

services/movie_session.py
```diff
@@ -0,0 +1,34 @@
...
+    return MovieSession.objects.all()
+
+
+def get_movie_session_by_id(movie_session_id):
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(session_id, show_time=None,
+                         movie_id=None, cinema_hall_id=None):
+    if show_time is not None:
```
SergeyDev1996, line 24: we could use just /IF show_time/, it will have the same meaning

SergeyDev1996 (approved):

services/movie_session.py
```diff
@@ -0,0 +1,34 @@
...
+        MovieSession.objects.filter(id=session_id).update(show_time=show_time)
+    if movie_id is not None:
+        MovieSession.objects.filter(id=session_id).update(movie=movie_id)
+    if cinema_hall_id is not None:
+        MovieSession.objects.filter(id=session_id)\
+            .update(cinema_hall=cinema_hall_id)
```
MasterpieceElbow, line 30: You don't need to `MovieSession.objects.filter` three times in this conditional block, you can do this before this block.

PR #43, opened 2022-06-21
Solution

darten360:


db/models.py
```diff
@@ -14,3 +14,34 @@ class Actor(models.Model):
 
     def __str__(self):
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
```
Flashmobber, line 23: Should be better if you added related_name for actor and genre

Flashmobber (changes requested):

services/movie.py
```diff
@@ -0,0 +1,50 @@
...
+        description=movie_description,
+    )
+
+    if genres_ids:
+        for i in genres_ids:
+            movie.genres.add(i)
+
+    if actors_ids:
+        for i in actors_ids:
+            movie.actors.add(i)
```
TPolina, line 50: same here

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+
+def create_movie_session(
+        movie_show_time: str,
+        movie_id: int,
+        cinema_hall_id: int
+):
+    MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie=Movie.objects.get(id=movie_id),
+        cinema_hall=CinemaHall.objects.get(id=cinema_hall_id)
```
TPolina, line 14: you can assign movie and cinema_hall through id and don't make unnecessary queries

TPolina (changes requested):

TPolina (approved):

PR #48, opened 2022-06-21
Solution

Evgeniy994:


services/movie.py
```diff
@@ -0,0 +1,39 @@
...
+            actors__id__in=actors_ids
+        )
+    elif genres_ids is not None and actors_ids is None:
+        return Movie.objects.filter(
+            genres__id__in=genres_ids
+        )
+    elif genres_ids is None and actors_ids is not None:
+        return Movie.objects.filter(
+            actors__id__in=actors_ids
+        )
```
katevasilenko, line 19: also, you can create this function similar as you create "create_movie"

katevasilenko (approved):

services/movie.py
MasterpieceElbow, line 19: It is a bad approach. Kate is right. Get all movies and then filter it for every argument passed.

MasterpieceElbow (changes requested):

PR #54, opened 2022-06-22
Solution

NOSSUDAK:


Ed12816 (approved):

db/models.py
```diff
@@ -14,3 +14,36 @@ class Actor(models.Model):
...
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
+
+    def __str__(self):
+        return f"{self.title}"
```
TPolina, line 26: you don't need an f-string here. `self.title` is a string already

db/models.py
```diff
@@ -14,3 +14,36 @@ class Actor(models.Model):
...
+
+
+class CinemaHall(models.Model):
+    name = models.CharField(max_length=255)
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
+
+    def __str__(self):
+        return f"{self.name}"
```
TPolina, line 35: same here

PR #58, opened 2022-06-27
'Solution'

YuraBeseda:


services/movie.py
```diff
@@ -0,0 +1,38 @@
...
+):
+    movies_queryset = Movie.objects.all()
+
+    if genres_ids:
+        movies_queryset = movies_queryset.filter(genres__id__in=genres_ids)
+
+    if actors_ids:
+        movies_queryset = movies_queryset.filter(actors__id__in=actors_ids)
+
+    return movies_queryset
```
vasylhnatiuk, line 16: "If both genres_ids and actors_ids are provided, the method returns movies, that have at least one genre from genres_ids and one actor from actors_ids" it needs one more option if get_movies takes 2 arguments

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

TPolina (approved):

PR #61, opened 2022-07-07
Solution

alix-mak:


services/movie.py
```diff
@@ -0,0 +1,32 @@
...
+
+def get_movie_by_id(movie_id):
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: List[int] = None,
+        actors_ids: List[int] = None
```
Abnormaltype, line 23: you can write `list[int]`, because we use Python 3.10

Abnormaltype (approved):

services/movie.py
alix-mak: ok ) my Python is 3.8 unfortunately. Glad to see you, Alex )

Abnormaltype: me too)

PR #63, opened 2022-07-07
Solution

drivemax26:


db/models.py
```diff
@@ -1,8 +1,56 @@
+# from django.db import models
```
IvanRamyk, line 1: Please, remove extra comments

IvanRamyk (changes requested):

db/models.py
```diff
@@ -2,7 +2,7 @@
 
 
 class Genre(models.Model):
-    name = models.CharField(max_length=255, unique=True)
+    name = models.CharField(max_length=255)
```
Abnormaltype, line 5: `name` should be `unique`.

Abnormaltype (changes requested):

Abnormaltype (changes requested):
also you should to add migrations.

Abnormaltype (approved):
well done!

PR #72, opened 2022-08-09
solution

tarkaiev:


services/cinema_hall.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+def create_cinema_hall(hall_name: str, hall_rows: int, hall_seats_in_row: int):
+    new_cinema_hall = CinemaHall.objects.create(
+        name=hall_name,
+        rows=hall_rows,
+        seats_in_row=hall_seats_in_row
+    )
+
+    return new_cinema_hall
```
Abnormaltype, line 15: this method should return nothing

services/movie.py
```diff
@@ -0,0 +1,39 @@
+from typing import List
+
+from db.models import Movie
+
+
+def get_movies(genres_ids: List[int] = None, actors_ids: List[int] = None):
```
Abnormaltype, line 6:
```suggestion
def get_movies(genres_ids: list[int] = None, actors_ids: list[int] = None):
```

use python 3.10+)

services/movie.py
```diff
@@ -0,0 +1,39 @@
...
+        description=movie_description
+    )
+
+    if genres_ids:
+        new_movie.genres.set(genres_ids)
+
+    if actors_ids:
+        new_movie.actors.set(actors_ids)
+
+    return new_movie
```
Abnormaltype, line 39: this method should return nothing

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+        movie_show_time: datetime,
+        movie_id: int,
+        cinema_hall_id: int
+):
+    new_movie_session = MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
+    return new_movie_session
```
Abnormaltype, line 16: this method should return nothing

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: datetime = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None,
+):
+    updated_session = MovieSession.objects.get(id=session_id)
```
Abnormaltype, line 35: use this function `get_movie_session_by_id`

Abnormaltype (changes requested):

services/movie.py
tarkaiev: finally found a reason to update from my 3.7

Abnormaltype: ))

Abnormaltype (approved):
Well done!

PR #101, opened 2022-10-06
Solution

wQuelS:


AriohBelskij (approved):
good!

db/models.py
```diff
@@ -4,13 +4,44 @@
...
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="actors")
+    genres = models.ManyToManyField(Genre, related_name="genres")
+
+    def __str__(self) -> str:
+        return f"{self.title}"
```
Nattalli, line 26: why in the `Genre` class you didn't use f-string for one word but here you are?

db/models.py
```diff
@@ -4,13 +4,44 @@
...
+class CinemaHall(models.Model):
+    name = models.CharField(max_length=255)
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
+
+    def capacity(self) -> int:
+        return self.rows * self.seats_in_row
+
+    def __str__(self) -> str:
+        return f"{self.name}"
```
Nattalli, line 38: the same here

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+        cinema_hall_id: int
+) -> None:
+    MovieSession.objects.create(
+        show_time=movie_show_time,
+        cinema_hall=CinemaHall.objects.get(id=cinema_hall_id),
+        movie=Movie.objects.get(id=movie_id)
+    )
+
+
+def get_movies_sessions(session_date: datetime = None) -> QuerySet:
```
Nattalli, line 20:
```
get_movies_sessions, takes optional string session_date ...
```
type must be `str`

Nattalli (changes requested):
Well done, left some comments

db/models.py
wQuelS: dunno, i'll fix it

Nattalli (approved):
Well done!

PR #102, opened 2022-10-07
Solution

AriohBelskij:


wQuelS (approved):

db/models.py
```diff
@@ -14,3 +14,34 @@ class Actor(models.Model):
...
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="actors")
+    genres = models.ManyToManyField(Genre, related_name="genres")
+
+    def __str__(self) -> str:
+        return f"{self.title}"
```
Nattalli, line 26: the same as `return self.title`?

Nattalli: I think that f-string is unnecessary here

db/models.py
```diff
@@ -14,3 +14,34 @@ class Actor(models.Model):
...
+
+
+class CinemaHall(models.Model):
+    name = models.CharField(max_length=255)
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
+
+    def __str__(self) -> str:
+        return f"{self.name}"
```
Nattalli, line 35: the same here

services/movie_session.py
```diff
@@ -0,0 +1,51 @@
...
+
+    if cinema_hall_id:
+        update_session.cinema_hall_id = cinema_hall_id
+
+    update_session.save()
+
+
+def delete_movie_session_by_id(session_id: int) \
+        -> MovieSession.objects:
+    return MovieSession.objects.filter(id=session_id).delete()
```
Nattalli, line 51: `.get()` method doesn't work here?

Nattalli (changes requested):
Summary: good job!
But left some comments

Nattalli (approved):
Well!

PR #108, opened 2022-10-11
Solution

mltdll:


AVyha (approved):
Looks good

db/models.py
```diff
@@ -14,3 +14,38 @@ class Actor(models.Model):
...
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="movies")
+    genres = models.ManyToManyField(Genre, related_name="movies")
+
+    def __str__(self) -> str:
+        return f"{self.title}"
```
Y-Havryliv, line 26: you don't need f-string here

db/models.py
```diff
@@ -14,3 +14,38 @@ class Actor(models.Model):
...
+class CinemaHall(models.Model):
+    name = models.CharField(max_length=255)
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
+
+    def capacity(self) -> int:
+        return self.rows * self.seats_in_row
+
+    def __str__(self) -> str:
+        return f"{self.name}"
```
Y-Havryliv, line 38: don't overuse f-string

db/models.py
```diff
@@ -14,3 +14,38 @@ class Actor(models.Model):
...
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(
+        CinemaHall, on_delete=models.CASCADE, related_name="movie_session"
```
Y-Havryliv, line 44: it would be better to name it `movie_sessions`

db/models.py
```diff
@@ -14,3 +14,38 @@ class Actor(models.Model):
...
+    )
+    movie = models.ForeignKey(
+        Movie, on_delete=models.CASCADE, related_name="movie_session"
```
Y-Havryliv, line 47: same here

services/cinema_hall.py
```diff
@@ -0,0 +1,15 @@
+from django.db.models import QuerySet
+
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> QuerySet:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(
```
Y-Havryliv, line 10: it's a good practice when your functions return smth

Y-Havryliv (changes requested):

services/cinema_hall.py
mltdll: I thought about it while implementing the functions, but the instructions didn't specify return value, so I assumed it to be `None`.

Y-Havryliv (approved):
Well done!

PR #109, opened 2022-10-13
Solution for the task

olia-trofymchuk:


db/models.py
```diff
@@ -14,3 +14,34 @@ class Actor(models.Model):
...
+        return self.rows * self.seats_in_row
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall, on_delete=models.CASCADE)
+    movie = models.ForeignKey(Movie, on_delete=models.CASCADE)
+
+    def __str__(self) -> str:
+        return f"{self.movie} {self.show_time.strftime('%Y-%m-%d %H:%M:%S')}"
```
mltdll, line 47: You can remove `.strftime()`, since this is default string format of a `datetime` object.

services/movie.py
```diff
@@ -0,0 +1,43 @@
+from django.db.models import QuerySet
+
+from db.models import Actor, Genre, Movie
+
+
+def get_movies(genres_ids: list[Genre] = None,
+               actors_ids: list[Actor] = None) -> QuerySet:
+    queryset = Movie.objects.all()
+
+    if genres_ids and actors_ids:
```
mltdll, line 10: My first impression is that you don't need this check. But if its purpose is to reduce the number of the database queries, then I guess it's fine.

services/movie_session.py
```diff
@@ -0,0 +1,51 @@
...
+
+    return movie
+
+
+def get_movies_sessions(session_date: datetime = None) -> QuerySet:
+    queryset = MovieSession.objects.all()
+
+    if session_date:
+        queryset = queryset.filter(show_time__date=session_date)
+        return queryset
```
mltdll, line 25: This `return` statement is redundant.

services/movie_session.py
```diff
@@ -0,0 +1,51 @@
...
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(session_id: int,
+                         show_time: datetime = None,
+                         movie_id: int = None,
+                         cinema_hall_id: int = None) -> None:
+    movie_session = MovieSession.objects.filter(id=session_id)
+
+    if show_time:
```
mltdll, line 40: If you are trying to reduce the number of queries then you should use Python object interface instead of `.update()` method and call `.save()` at the end of the function.

mltdll (changes requested):

db/models.py
olia-trofymchuk: ok, thank you :)

services/movie_session.py
olia-trofymchuk: okaay, i'll implement it. thank u a lot!

mltdll (approved):

db/models.py
```diff
@@ -14,3 +14,34 @@ class Actor(models.Model):
...
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
+
+    def __str__(self) -> str:
+        return f"{self.title}"
```
Y-Havryliv, line 26: do you need f-string here?

db/models.py
```diff
@@ -14,3 +14,34 @@ class Actor(models.Model):
...
+
+
+class CinemaHall(models.Model):
+    name = models.CharField(max_length=255)
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
+
+    def __str__(self) -> str:
+        return f"{self.name}"
```
Y-Havryliv, line 35: don't overuse f-string

services/cinema_hall.py
```diff
@@ -0,0 +1,17 @@
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> CinemaHall:
```
Y-Havryliv, line 4: it returns `QuerySet` actually

services/movie_session.py
```diff
@@ -0,0 +1,52 @@
...
+
+from db.models import MovieSession, CinemaHall, Movie
+
+import datetime
+
+
+def create_movie_session(movie_show_time: datetime,
+                         movie_id: int,
+                         cinema_hall_id: int) -> Movie:
+    movie = MovieSession.objects.create(
```
Y-Havryliv, line 11: it's not `movie`, it's `movie_session`)

services/movie_session.py
```diff
@@ -0,0 +1,52 @@
...
+        show_time=movie_show_time,
+        cinema_hall=CinemaHall.objects.get(id=cinema_hall_id),
```
Y-Havryliv, line 13: you don't need to get `CinemaHall` and `Movie`, just use the ids

services/movie_session.py
```diff
@@ -0,0 +1,52 @@
...
+        movie_session.movie_id = movie_id
+
+    if cinema_hall_id:
+        movie_session.cinema_hall_id = cinema_hall_id
+
+    movie_session.save()
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    MovieSession.objects.filter(id=session_id).delete()
```
Y-Havryliv, line 52: use `get` to get 1 object

Y-Havryliv (changes requested):

Y-Havryliv (approved):
Well done!

Y-Havryliv (approved):

PR #110, opened 2022-10-15
Solution

yevhenii-nevmyvako:


services/movie_session.py
```diff
@@ -0,0 +1,49 @@
...
+                         show_time: int = None,
+                         movie_id: int = None,
+                         cinema_hall_id: int = None
+                         ) -> None:
+    if show_time:
+        MovieSession.objects.filter(id=session_id).update(show_time=show_time)
+    if movie_id:
+        MovieSession.objects.all().update(movie_id=movie_id)
+    if cinema_hall_id:
+        MovieSession.objects.all().update(cinema_hall_id=cinema_hall_id)
```
DmytroSvirsa, line 45: let's remove code duplication of MovieSession.objects by providing variable and using get when working with single object

DmytroSvirsa (changes requested):

services/movie_session.py
yevhenii-nevmyvako: i don`t undarstand how can use get() with update() in this task, i have a mistake -'MovieSession' object has no attribute 'update'

yevhenii-nevmyvako: chenge all mistakes

DmytroSvirsa (approved):

PR #111, opened 2022-10-15
Solution

IvanStored:


AriohBelskij (approved):

services/movie_session.py
```diff
@@ -0,0 +1,58 @@
...
+from db.models import MovieSession, Movie, CinemaHall
+
+
+def create_movie_session(
+        movie_show_time: datetime,
+        movie_id: int,
+        cinema_hall_id: int
+) -> None:
+    movie = Movie.objects.get(id=movie_id)
+    cinema_hall = CinemaHall.objects.get(id=cinema_hall_id)
```
TPolina, line 14: you don't need to perform these queries. You can set `cinema_hall` and `movie` by `id`s and not only objects

TPolina (changes requested):

services/movie_session.py
```diff
@@ -0,0 +1,56 @@
...
+
+    if cinema_hall_id:
+        session.cinema_hall_id = cinema_hall_id
+
+    session.save()
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    movie_session = MovieSession.objects.get(id=session_id)
+    movie_session.delete()
```
MasterpieceElbow, line 56: You can delete it simply in one query.

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #113, opened 2022-10-16
Solution

luxuriant777:


db/models.py
```diff
@@ -14,3 +14,42 @@ class Actor(models.Model):
...
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField(blank=True, null=True)
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
+
+    def __str__(self) -> str:
+        return f"{self.title}"
```
mltdll, line 26: You don't need f-string here, this value is already a string.
Also at $35.

services/movie.py
```diff
@@ -0,0 +1,40 @@
...
+def get_movie_by_id(movie_id: int) -> Movie.objects:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
+) -> Movie.objects:
```
mltdll, line 28: This function returns `Movie`, not `Movie.objects`.
The other functions as well.

services/movie_session.py
```diff
@@ -0,0 +1,57 @@
...
+
+from services.cinema_hall import get_cinema_hall_by_id
+
+
+def create_movie_session(
+        movie_show_time: str,
+        movie_id: int,
+        cinema_hall_id: int
+) -> MovieSession.objects:
+    new_session = MovieSession.objects.create(
```
mltdll, line 13: You can safely put `return` right here, without introducing a variable.

mltdll (changes requested):

db/models.py
luxuriant777: Thanks. Changed

mltdll (approved):

services/movie.py
luxuriant777: Thanks. Changed

services/movie_session.py
luxuriant777: Thanks. Changed

MasterpieceElbow (approved):
GJ!

PR #118, opened 2022-10-19
Solution

taras-andruschenko:


db/models.py
```diff
@@ -14,3 +14,42 @@ class Actor(models.Model):
...
+    def __str__(self) -> str:
+        return f"{self.name}"
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(
+        CinemaHall,
+        on_delete=models.CASCADE,
+        related_name="movie_session"
```
mltdll, line 46: "Many" parts of one-to-many and many-to-many relationships should have plural forms as `related_name`s.
E.g. `movie_sessions`.

db/models.py
```diff
@@ -14,3 +14,42 @@ class Actor(models.Model):
...
+    )
+    movie = models.ForeignKey(
+        Movie,
+        on_delete=models.CASCADE,
+        related_name="movie_session"
```
mltdll, line 51: This one as well.

services/movie_session.py
```diff
@@ -0,0 +1,60 @@
+from django.db.models import QuerySet
+
+from db.models import MovieSession, CinemaHall, Movie
+
+
+def create_movie_session(
+        movie_show_time: str,
+        movie_id: int,
+        cinema_hall_id: int
+) -> None:
```
mltdll, line 10: It's a good practice when your functions return something. :)

mltdll (changes requested):

mltdll (approved):

db/models.py
```diff
@@ -14,3 +14,42 @@ class Actor(models.Model):
...
+class CinemaHall(models.Model):
+    name = models.CharField(max_length=255)
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
+
+    def capacity(self) -> int:
+        return self.rows * self.seats_in_row
+
+    def __str__(self) -> str:
+        return f"{self.name}"
```
mltdll, line 38: This one as well.

mltdll (changes requested):

db/models.py
```diff
@@ -14,3 +14,42 @@ class Actor(models.Model):
...
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
+
+    def __str__(self) -> str:
+        return f"{self.title}"
```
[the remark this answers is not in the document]
taras-andruschenko: Agree

mltdll (approved):

services/movie_session.py
```diff
@@ -0,0 +1,60 @@
...
+
+def create_movie_session(
+        movie_show_time: str,
+        movie_id: int,
+        cinema_hall_id: int
+) -> MovieSession:
+    return MovieSession.objects.create(
+        show_time=movie_show_time,
+        cinema_hall=CinemaHall.objects.get(id=cinema_hall_id),
+        movie=Movie.objects.get(id=movie_id)
```
TPolina, line 14: you don't need to perform these queries. You can set cinema_hall and movie by ids and not only objects

TPolina (changes requested):

services/movie_session.py
taras-andruschenko: fixed, but I still don't understand, why I have to use `movie_session_id` with only one underline... Could you suggest me a topic what I probably missed

MasterpieceElbow (approved):

PR #119, opened 2022-10-19
Solution

Aleksandro777:


services/cinema_hall.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+def get_cinema_halls() -> QuerySet:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(
+        hall_name: str,
+        hall_rows: int,
+        hall_seats_in_row: int) -> None:
```
mltdll, line 12: It's better to return something from functions like this one.

services/movie.py
```diff
@@ -0,0 +1,44 @@
...
+def get_movie_by_id(movie_id: int) -> QuerySet:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
+) -> QuerySet:
```
mltdll, line 32: This function seems to return `Movie` object, not `QuerySet`.

mltdll (changes requested):

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
...
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
+    if session_date:
+        return MovieSession.objects.filter(show_time__date=session_date)
+    return MovieSession.objects.all()
+
+
+def get_movie_session_by_id(movie_session_id: int) -> QuerySet:
```
mltdll, line 27: This function also doesn't seem to return `QuerySet`.

services/movie.py
```diff
@@ -0,0 +1,44 @@
...
+
+    if genres_ids:
+        queryset = queryset.filter(
+            genres__id__in=genres_ids,
+        )
+
+    return queryset
+
+
+def get_movie_by_id(movie_id: int) -> QuerySet:
```
mltdll, line 23: Sorry, missed this one: this function also create a `Movie` instance, not `QuerySet`.

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
+from datetime import datetime
+from django.db.models import QuerySet
+from db.models import MovieSession, Movie, CinemaHall
+
+
+def create_movie_session(
+        movie_show_time: datetime,
+        movie_id: int,
+        cinema_hall_id: int
+) -> None:
```
mltdll, line 10: You may also change functions `create_movie_session()`, `update_movie_session()`, and `delete_movie_session()` so that they return created/modified/deleted objects as well.

mltdll (changes requested):

services/movie_session.py
```diff
@@ -47,7 +47,9 @@ def update_movie_session(
 
     new_session.save()
 
+    return new_session
 
-def delete_movie_session_by_id(session_id: int) -> None:
+
+def delete_movie_session_by_id(session_id: int) -> MovieSession.objects:
```
mltdll, line 53: `MovieSession.objects` is a manager object. `.delete()` method returns a `tuple` of `int` and `dict`.

mltdll (changes requested):

mltdll (approved):

services/movie_session.py
```diff
@@ -0,0 +1,55 @@
...
+from db.models import MovieSession, Movie, CinemaHall
+
+
+def create_movie_session(
+        movie_show_time: datetime,
+        movie_id: int,
+        cinema_hall_id: int
+) -> MovieSession:
+    movie = Movie.objects.get(id=movie_id)
+    cinema_hall = CinemaHall.objects.get(id=cinema_hall_id)
```
Nattalli, line 12: read 1 here:

Nattalli: optimize it

services/movie_session.py
```diff
@@ -0,0 +1,55 @@
...
+
+    if cinema_hall_id:
+        new_session.cinema_hall_id = cinema_hall_id
+
+    new_session.save()
+
+    return new_session
+
+
+def delete_movie_session_by_id(session_id: int) -> tuple[int, dict]:
```
Nattalli, line 53: you can return `None`
in the function just delete object and don't use `return` statement

Nattalli (changes requested):

services/movie_session.py
Nattalli: ohm I think it is optional

Nattalli: Yes, the return is optional in this function. I just thought that this function had a `None` type earlier and you asked to change it. But now I see that you were right

Снимок экрана 2022-10-24 в 21 33 50

Aleksandro777: So we leave the return here?

Nattalli (approved):
well done!

PR #126, opened 2022-10-22
Solution

LiudmylaKulinchenko:


db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
...
+        return self.rows * self.seats_in_row
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall, on_delete=models.CASCADE)
+    movie = models.ForeignKey(Movie, on_delete=models.CASCADE)
+
+    def __str__(self) -> str:
+        movie_time = self.show_time.strftime("%Y-%m-%d %H:%M:%S")
```
mltdll, line 47: This line can be removed, since this is the default string representation of datetime object (one without fractional seconds).

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: datetime.date = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> None:
+    movie_session = MovieSession.objects.all().filter(id=session_id)
```
mltdll, line 39: `.all()` is not necessary here.

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+
+    if show_time:
+        movie_session.update(show_time=show_time)
```
mltdll, line 42: It would be more efficient in terms of database interactions to get a `MovieSession` object, set required fields via Python object interface, and call `obj.save()` afterwards.

mltdll (changes requested):

mltdll (approved):

MasterpieceElbow (approved):
GJ!

PR #127, opened 2022-10-22
Solution

VadymShkarbul:


db/models.py
```diff
@@ -14,3 +14,34 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
```
Yaroslav-Stetsko: I suggest adding related_name since its best practice

Yaroslav-Stetsko (approved):
Well done!

services/cinema_hall.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def get_cinema_halls() -> QuerySet:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(hall_name: str,
+                       hall_rows: int,
+                       hall_seats_in_row: int) -> None:
+    if not CinemaHall.objects.filter(name=hall_name).exists():
```
Nattalli, line 13: this condition is unnecessary

services/cinema_hall.py
```diff
@@ -0,0 +1,22 @@
...
+        CinemaHall.objects.create(
+            name=hall_name,
+            rows=hall_rows,
+            seats_in_row=hall_seats_in_row
+        )
+
+
+def del_cinema_hall_by_title(hall_name: str) -> None:
+    CinemaHall.objects.filter(name=hall_name).delete()
```
Nattalli, line 22: this method is not mentioned in the task

services/movie.py
```diff
@@ -0,0 +1,43 @@
...
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(movie_title: str,
+                 movie_description: str,
+                 genres_ids: list = None,
+                 actors_ids: list = None) -> None:
+    if not Movie.objects.filter(title=movie_title).exists():
```
Nattalli, line 27: also unnecessary condition

services/movie.py
```diff
@@ -0,0 +1,43 @@
...
+            new_movie.genres.set(genres_ids)
+
+        if actors_ids:
+            new_movie.actors.set(actors_ids)
+
+
+def del_movie_by_title(movie_title: str) -> None:
+    Movie.objects.filter(
+        title=movie_title
+    ).delete()
```
Nattalli, line 43: also not mentioned in the task

services/movie_session.py
```diff
@@ -0,0 +1,47 @@
...
+                         movie_id: int,
+                         cinema_hall_id: int) -> None:
+    MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
```
Nattalli, line 18: -> `MovieSession`

services/movie_session.py
```diff
@@ -0,0 +1,47 @@
...
+    if session_date:
+        return MovieSession.objects.filter(
+            show_time__date=datetime.strptime(session_date, "%Y-%m-%d").date())
```
Nattalli, line 21: ```show_time__date=session_date```
will work in the same way

services/movie_session.py
```diff
@@ -0,0 +1,47 @@
...
+
+def update_movie_session(
+        session_id: int,
+        show_time: str = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None) -> None:
+
+    movie_session = MovieSession.objects.filter(id=session_id)
+
+    if movie_session.exists():
```
Nattalli, line 33: unnecessary condition

services/movie_session.py
```diff
@@ -0,0 +1,47 @@
...
+        if cinema_hall_id:
+            movie_session.update(cinema_hall=cinema_hall_id)
+
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    MovieSession.objects.filter(id=session_id).delete()
```
Nattalli, line 47: read the 3 item here

settings.py
```diff
@@ -21,6 +21,6 @@
 
 USE_I18N = True
 
-USE_TZ = True
+USE_TZ = False
 
-INSTALLED_APPS = ("db",)
+INSTALLED_APPS = ("db", "services")
```
Nattalli, line 26: why did you do it?

Nattalli (changes requested):

services/cinema_hall.py
VadymShkarbul: I used it for personal purposes and forgot to delete it

services/movie_session.py
VadymShkarbul: Yes, but I learned about it only in the next episode of "Danilo and Django"  :)

services/movie_session.py
VadymShkarbul: I was sure that both all() and filter() return QuerySet or am I wrong?

services/movie_session.py
VadymShkarbul: I was thinking that if the string with the date is not formatted as in the table, should we reformat it or does Django format the dates?

VadymShkarbul: Yes. You are totally right!

services/movie_session.py
VadymShkarbul: Yes, thanks! I even fixed it, but did not commited! Still need to read the checklist before the review. Guilty!

services/movie_session.py
Nattalli: you are right

Nattalli (approved):
👍🏻

PR #130, opened 2022-10-22
Solution

misha-doroshenko:


db/models.py
```diff
@@ -14,3 +14,42 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="actors_set")
+    genres = models.ManyToManyField(Genre, related_name="genres_set")
```
mltdll, line 23: I'm not really sure about this, but wouldn't it be better to set `related_name` to a simple plural form? e. g. `actors`, `genres`.

db/models.py
```diff
@@ -14,3 +14,42 @@ class Actor(models.Model):
...
+    def __str__(self) -> str:
+        return self.name
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(
+        CinemaHall,
+        on_delete=models.CASCADE,
+        related_name="halls"
```
mltdll, line 46: I think it's more intuitive to seek this related name as `cinema_halls` when you're working with the model. Therefore I'd suggest to use this name.

db/models.py
```diff
@@ -14,3 +14,42 @@ class Actor(models.Model):
...
+    )
+    movie = models.ForeignKey(
+        Movie,
+        on_delete=models.CASCADE,
+        related_name="movies"
+    )
+
+    def __str__(self) -> str:
+        return self.show_time.strftime(f"{self.movie.title} %Y-%m-%d %H:%M:%S")
```
mltdll, line 55: This returned value can be rewritten as `f"{self.movie.title} {self.show_time}"`.

services/movie.py
```diff
@@ -0,0 +1,41 @@
...
+    if genres_ids:
+        queryset = queryset.filter(genres__id__in=genres_ids)
+
+    if actors_ids:
+        queryset = queryset.filter(actors__id__in=actors_ids)
+
+    return queryset
+
+
+def get_movie_by_id(movie_id: int) -> Any:
```
mltdll, line 20: You mean `-> Movie`?

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
...
+def get_movies_sessions(session_date: str = None) -> QuerySet:
+    queryset = MovieSession.objects.all()
+
+    if session_date:
+        queryset = queryset.filter(show_time__date=session_date)
+
+    return queryset
+
+
+def get_movie_session_by_id(movie_session_id: int) -> Any:
```
mltdll, line 16: `-> MovieSession`

mltdll (changes requested):

mltdll (approved):

MasterpieceElbow (approved):
GJ!

PR #133, opened 2022-10-23
Solution

HStakhiv:


mltdll (approved):

services/cinema_hall.py
```diff
@@ -0,0 +1,17 @@
...
+
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> QuerySet:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(
+    hall_name: str, hall_rows: str, hall_seats_in_row: int
```
Nattalli, line 11: btw, why `hall_rows` is an `str` here and `IntegerField` in the model?

HStakhiv: Stupid mistake. I copied `hall_name: str` and replaced `name` with `rows` and not the types annotation

PR #134, opened 2022-10-23
Solution

KatrinLazarenko:


services/cinema_hall.py
```diff
@@ -0,0 +1,17 @@
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> CinemaHall.objects:
```
HStakhiv, line 4: I think it is better to `import` and use `QuerySet` instead of `CinemaHall.objects`

services/movie.py
```diff
@@ -0,0 +1,33 @@
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
+) -> Movie.objects:
```
HStakhiv, line 7: I think it is better to `import` and use `QuerySet` instead of `Movie.objects`

services/movie_session.py
```diff
@@ -0,0 +1,44 @@
...
+        cinema_hall_id: int
+) -> None:
+    MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie=Movie.objects.get(id=movie_id),
+        cinema_hall=CinemaHall.objects.get(id=cinema_hall_id)
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> MovieSession.objects:
```
HStakhiv, line 16: I think it is better to `import` and use `QuerySet` instead of `MovieSession.objects`

HStakhiv (changes requested):

HStakhiv (approved):
Looks good!

services/cinema_hall.py
KatrinLazarenko: Thanks for your advice.

MasterpieceElbow (approved):
GJ!

PR #135, opened 2022-10-23
Develop

matechris:


mltdll (approved):

db/models.py
```diff
@@ -14,3 +14,42 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
```
Nattalli, line 23: add `related_name`

services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+                         movie_id: int,
+                         cinema_hall_id: id) -> MovieSession:
+
+    new_movie_session = MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
+
+    return new_movie_session
```
Nattalli, line 18: btw, you can make 
`return MovieSession.objects.create(...)`

Nattalli: but it's optional

Nattalli (approved):
Well done!

PR #138, opened 2022-10-24
Solution

skrypka-dmytro:


db/models.py
```diff
@@ -14,3 +14,40 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="actors")
+    genres = models.ManyToManyField(Genre, related_name="genre")
```
mltdll, line 23: Related name should probably be `genres`.

services/cinema_hall.py
```diff
@@ -0,0 +1,15 @@
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> CinemaHall.objects:
```
mltdll, line 4: `CinemaHall.objects` is a manager object. This function does not return it.
Fix it everywhere.

services/movie_session.py
```diff
@@ -0,0 +1,62 @@
...
+
+def create_movie_session(
+        movie_show_time: str,
+        movie_id: int,
+        cinema_hall_id: int
+) -> None:
+    MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie=Movie.objects.get(id=movie_id),
+        cinema_hall=CinemaHall.objects.get(id=cinema_hall_id),
```
mltdll, line 14: These two queries are not necessary: you can set foreign key directly by id: `movie_id`.

services/movie_session.py
```diff
@@ -0,0 +1,62 @@
...
+    )
+
+
+def get_movies_sessions(
+        session_date: str = None
+) -> MovieSession.objects:
+    queryset = MovieSession.objects.all()
+
+    if session_date is not None:
+        queryset = MovieSession.objects.filter(
```
mltdll, line 24: This function features unnecessary database calls. It can be improved in one of the following ways:
```python
if condition:
    return Model.objects.filter(...)
return Model.objects.all()
```
```python
qs = Model.objects.all()
if condition:
    qs = qs.filter(...)
return qs
```

services/movie_session.py
```diff
@@ -0,0 +1,62 @@
...
+            show_time__date=session_date
+        )
+
+    return queryset
+
+
+def get_movie_session_by_id(
+        movie_session_id: int
+) -> MovieSession.objects:
+    return MovieSession.objects.get(
```
mltdll, line 34: It looks like there's some unnecessary line breaks here.

mltdll (changes requested):

services/movie_session.py
skrypka-dmytro: what is the difference

mltdll: Oh, wait. Your implementation actually doesn't call database twice.
Either way, it's a bad practice to put something like a queryset into a variable and not reuse it.

mltdll: The difference is that in your solution, the function creates a queryset and immediately throws it away if `session_date` is not `None`. There's no reason to do it like that. You should either continue to use this queryset or not create it at all.

services/movie.py
```diff
@@ -0,0 +1,36 @@
+from db.models import Movie
+
+
+def get_movies(genres_ids: list[int] = None,
+               actors_ids: list[int] = None) -> Movie.objects:
```
mltdll, line 5: There are incorrect return type annotations in this file as well.

mltdll (changes requested):

mltdll:
Good think I'm not checking solutions for flake8 errors.

services/movie.py
```diff
@@ -0,0 +1,39 @@
...
+    if genres_ids is not None:
+        queryset = queryset.filter(genres__id__in=genres_ids)
+
+    if actors_ids is not None:
+        queryset = queryset.filter(actors__id__in=actors_ids)
+
+    return queryset
+
+
+def get_movie_by_id(movie_id: int = None) -> Movie.objects:
```
Nattalli, line 20: `movie_id` is required

Nattalli: and should this function just return `Movie`?

Nattalli (changes requested):

services/movie.py
```diff
@@ -17,7 +17,7 @@ def get_movies(genres_ids: list[int] = None,
     return queryset
 
 
-def get_movie_by_id(movie_id: int = None) -> Movie.objects:
+def get_movie_by_id(movie_id: int = None) -> Movie:
```
Nattalli, line 20: okay but still, remove `None`
as `movie_is` is required
```python
def get_movie_by_id(movie_id: int) -> Movie:
```

Nattalli (changes requested):

Nattalli (approved):
👍🏻

PR #142, opened 2022-10-24
Solution

graandda:


services/movie_session.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+from db.models import MovieSession, Movie, CinemaHall
+
+
+def create_movie_session(movie_show_time: str, movie_id: int,
+                         cinema_hall_id: int) -> None:
+    MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie=Movie.objects.get(id=movie_id),
```
FranchukS, line 11: You can set the id number without getting an instance (when you do, you load DB again)
Like: movie_id = movie_id

Nattalli (changes requested):

db/models.py
```diff
@@ -19,8 +19,8 @@ def __str__(self) -> str:
 class Movie(models.Model):
     title = models.CharField(max_length=255)
     description = models.TextField()
-    actors = models.ManyToManyField(Actor)
-    genres = models.ManyToManyField(Genre)
+    actors = models.ManyToManyField(Actor, related_name="actors")
+    genres = models.ManyToManyField(Genre, related_name="genres")
```
Nattalli, line 23: btw, that are bad names: `movies` would be better. Read the checklist!!
https://github.com/mate-academy/py-movie-moviesession-cinemahall/blob/master/checklist.md

Nattalli (changes requested):

Y-Havryliv (approved):
Well done!

PR #143, opened 2022-10-24
Solved the task

Cerne13:


services/movie_sessions.py
```diff
@@ -0,0 +1,53 @@
...
+    return MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie=Movie.objects.get(id=movie_id),
+        cinema_hall=CinemaHall.objects.get(id=cinema_hall_id)
+    )
+
+
+def get_movies_sessions(
+    session_date: str = None
+) -> QuerySet(MovieSession):
```
mltdll, line 20: Square brackets should be used here.

services/cinema_hall.py
```diff
@@ -0,0 +1,19 @@
+from django.db.models import QuerySet
+
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> QuerySet:
```
mltdll, line 6: You add more details to return type annotations in other places, so why don't you do the same here, for the sake of consistency? :)

mltdll (changes requested):

services/movie_sessions.py
Cerne13: Oh, yes. Somehow overlooked this. Thank you :)

services/cinema_hall.py
Cerne13: Aha, went a bit lazy here :)

services/movie_sessions.py
```diff
@@ -0,0 +1,53 @@
...
+
+def create_movie_session(
+        movie_show_time: str,
+        movie_id: int,
+        cinema_hall_id: int
+) -> MovieSession:
+    return MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie=Movie.objects.get(id=movie_id),
+        cinema_hall=CinemaHall.objects.get(id=cinema_hall_id)
```
mltdll, line 14: Oops, I missed this one: you can pass ids directly to foreign key fields — use `fieldname_id` arguments. This will reduce the amount of database queries.

mltdll (approved):
Perfect! Unless I missed something else.

Cerne13:
Thank you! Your code reviews are the best!

tests/test_main.py
```diff
@@ -7,7 +7,7 @@
 from db.models import Actor, Genre, Movie, MovieSession, CinemaHall
 from services.movie import get_movies, get_movie_by_id, create_movie
 from services.cinema_hall import get_cinema_halls, create_cinema_hall
-from services.movie_session import (
+from services.movie_sessions import (
```
mltdll, line 10: Shouldn't this module be named `movie_session`?

mltdll (changes requested):
... Which I might have.

tests/test_main.py
Cerne13: Oh, you're right. Can't remember exactly why I've renamed it.
Fixed 👍

mltdll (approved):

MasterpieceElbow (approved):
GJ!

PR #144, opened 2022-10-24
Solution

FranchukS:


services/movie_session.py
```diff
@@ -0,0 +1,48 @@
...
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: datetime = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> None:
```
Mowka7, line 35: Are you sure that this method should return nothing?

Mowka7 (changes requested):

services/movie_session.py
```diff
@@ -0,0 +1,48 @@
...
+    session = MovieSession.objects.filter(id=session_id)
+    if show_time:
+        session.update(show_time=show_time)
+
+    if movie_id:
+        session.update(movie=movie_id)
+
+    if cinema_hall_id:
+        session.update(cinema_hall=cinema_hall_id)
```
Mowka7, line 44: I think you need to use `session.save()`

Mowka7 (changes requested):

services/movie.py
```diff
@@ -0,0 +1,37 @@
...
+    if genres_ids:
+        queryset = queryset.filter(genres__id__in=genres_ids)
+    if actors_ids:
+        queryset = queryset.filter(actors__id__in=actors_ids)
+
+    return queryset
+
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    if Movie.objects.filter(id=movie_id):
```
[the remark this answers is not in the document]
FranchukS: Yeah, you're right, I just thought to add it to avoid an error)

services/movie_session.py
FranchukS: Not in this case - I used ORM manager (objects) for update - it added changes itself.
You need to use save() when trying to add changes through the class instance:

session = MovieSession.objects.get(id=session_id) - become instance
session.show_time = show_time - add changes
session.save( )

FranchukS: Not in this case - I used ORM manager (objects) for update - it added changes itself.
You need to use save() when trying to add changes through the class instance:

session = MovieSession.objects.get(id=session_id) - become instance
session.show_time = show_time - add changes
session.save( )

Mowka7 (approved):
Well done!(〜￣▽￣)〜

MasterpieceElbow (approved):
GJ!

PR #145, opened 2022-10-24
Created classes models and service functions with many to many db man…

alexzubkoff:
…ipulations

### DIFF
--- a/db/models.py
+++ b/db/models.py
@@ -14,3 +14,34 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
+
+    def __str__(self) -> str:
+        return f"{self.title}"
+
+
+class CinemaHall(models.Model):
+    name = models.CharField(max_length=255)
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
+
+    def __str__(self) -> str:
+        return f"{self.name}"
+
+    def capacity(self) -> int:
+        return self.rows * self.seats_in_row
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall, on_delete=models.CASCADE)
+    movie = models.ForeignKey(Movie, on_delete=models.CASCADE)
+
+    def __str__(self) -> str:
+        return f"{self.movie} {self.show_time}"
--- a/services/cinema_hall.py
+++ b/services/cinema_hall.py
@@ -1,0 +1,12 @@
+from django.db.models import QuerySet
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> QuerySet:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(hall_name: str, hall_rows: int,
+                       hall_seats_in_row: int) -> None:
+    CinemaHall.objects.create(name=hall_name, rows=hall_rows,
+                              seats_in_row=hall_seats_in_row)
--- a/services/movie.py
+++ b/services/movie.py
@@ -1,0 +1,40 @@
+from django.db.models import QuerySet
+from db.models import Movie
+
+
+def get_movies(genres_ids: list[int] = None,
+               actors_ids: list[int] = None) -> QuerySet:
+    if not genres_ids and not actors_ids:
+        return Movie.objects.all()
+
+    if genres_ids and actors_ids:
+        return Movie.objects.filter(genres__id__in=genres_ids).\
+            filter(actors__id__in=actors_ids)
+
+    if genres_ids and not actors_ids:
+        return Movie.objects.filter(genres__id__in=genres_ids)
+
+    if actors_ids and not genres_ids:
+        return Movie.objects.filter(actors__id__in=actors_ids)
+
+
+def get_movie_by_id(movie_id: int) -> QuerySet:
+    if movie_id:
+        if Movie.objects.filter(id=movie_id):
+            return Movie.objects.get(id=movie_id)
+
+
+def create_movie(movie_title: str, movie_description: str,
+                 genres_ids: list[int] = None,
+                 actors_ids: list[int] = None) -> None:
+    if movie_title and movie_description:
+        movie = Movie.objects.create(title=movie_title,
+                                     description=movie_description)
+
+    if genres_ids:
+        for genre_id in genres_ids:
+            movie.genres.add(genre_id)
+
+    if actors_ids:
+        for actor_id in actors_ids:
+            movie.actors.add(actor_id)
--- a/services/movie_session.py
+++ b/services/movie_session.py
@@ -1,0 +1,49 @@
+from django.db.models import QuerySet
+from db.models import Movie, CinemaHall, MovieSession
+
+
+def create_movie_session(movie_show_time: str,
+                         movie_id: int,
+                         cinema_hall_id: int) -> None:
+    if movie_show_time and movie_id and cinema_hall_id:
+        cinema_hall = CinemaHall.objects.get(id=cinema_hall_id)
+        movie = Movie.objects.get(id=movie_id)
+        movie_session = MovieSession(
+            show_time=movie_show_time,
+            cinema_hall=cinema_hall,
+            movie=movie
+        )
+        movie_session.save()
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
+    query_set = MovieSession.objects.all()
+    if session_date:
+        query_set = query_set.filter(show_time__date=session_date)
+    return query_set
+
+
+def get_movie_session_by_id(movie_session_id: id) -> QuerySet:
+    if movie_session_id:
+        return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(session_id: int, show_time: str = None,
+                         movie_id: int = None,
+                         cinema_hall_id: int = None) -> None:
+    if session_id:
+        movie_session = MovieSession.objects.get(id=session_id)
+    if show_time:
+        movie_session.show_time = show_time
+    if movie_id:
+        movie = Movie.objects.get(id=movie_id)
+        movie_session.movie = movie
+    if cinema_hall_id:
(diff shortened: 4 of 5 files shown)


services/cinema_hall.py
```diff
@@ -0,0 +1,12 @@
+from django.db.models import QuerySet
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> QuerySet:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(hall_name: str, hall_rows: int,
+                       hall_seats_in_row: int) -> None:
```
serviza, line 10: The same moments like were in my code. Function should not return None.
It's better to return object.

services/movie.py
```diff
@@ -0,0 +1,40 @@
...
+            filter(actors__id__in=actors_ids)
+
+    if genres_ids and not actors_ids:
+        return Movie.objects.filter(genres__id__in=genres_ids)
+
+    if actors_ids and not genres_ids:
+        return Movie.objects.filter(actors__id__in=actors_ids)
+
+
+def get_movie_by_id(movie_id: int) -> QuerySet:
```
serviza, line 21: The same here. It will return one object

services/movie.py
```diff
@@ -0,0 +1,40 @@
...
+    if movie_id:
+        if Movie.objects.filter(id=movie_id):
+            return Movie.objects.get(id=movie_id)
+
+
+def create_movie(movie_title: str, movie_description: str,
+                 genres_ids: list[int] = None,
+                 actors_ids: list[int] = None) -> None:
```
serviza, line 29: The same here. Function will return an object.

services/movie_session.py
```diff
@@ -0,0 +1,49 @@
+from django.db.models import QuerySet
+from db.models import Movie, CinemaHall, MovieSession
+
+
+def create_movie_session(movie_show_time: str,
```
serviza, line 5: Type of show_time should be in datatype format

services/movie_session.py
```diff
@@ -0,0 +1,49 @@
...
+    if movie_id:
+        movie = Movie.objects.get(id=movie_id)
+        movie_session.movie = movie
+    if cinema_hall_id:
+        cinema_hall = CinemaHall.objects.get(id=cinema_hall_id)
+        movie_session.cinema_hall = cinema_hall
+    movie_session.save()
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
```
serviza, line 47: Check this annotation too

serviza (changes requested):

alexzubkoff:
Ok, thanks!

Пн, 24 окт. 2022 г. в 21:19, Sergiy Zakharenko ***@***.***>:

serviza (approved):
Cool!

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
...
+                         cinema_hall_id: int) -> MovieSession:
+    if movie_show_time and movie_id and cinema_hall_id:
+        cinema_hall = CinemaHall.objects.get(id=cinema_hall_id)
+        movie = Movie.objects.get(id=movie_id)
+        movie_session = MovieSession(
+            show_time=movie_show_time,
+            cinema_hall=cinema_hall,
+            movie=movie
+        )
+        movie_session.save()
```
Nattalli, line 17: better use `MovieSession .objects.create` instead of 
```python
a = MovieSession()
a.save()
```

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
...
+        return movie_session
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet | MovieSession:
```
Nattalli, line 21: `-> Queryset`
Remember:
`.all()`, `.filter()` - returns QuerySet
`.get()`, `.create()`, `.delete()` - return object class, `MovieSession` for example

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
...
+    query_set = MovieSession.objects.all()
+    if session_date:
+        query_set = query_set.filter(show_time__date=session_date)
+    return query_set
+
+
+def get_movie_session_by_id(movie_session_id: id) -> MovieSession:
+    if movie_session_id:
```
Nattalli, line 29: unnecessary condition

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
...
+        movie_session = MovieSession.objects.get(id=movie_session_id)
+        return movie_session
+
+
+def update_movie_session(session_id: int, show_time: str = None,
+                         movie_id: int = None,
+                         cinema_hall_id: int = None) -> MovieSession:
+    if session_id:
```
Nattalli, line 37: this condition is unnecessary

Nattalli: as it is required param

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
...
+
+def get_movie_session_by_id(movie_session_id: id) -> MovieSession:
+    if movie_session_id:
+        movie_session = MovieSession.objects.get(id=movie_session_id)
+        return movie_session
+
+
+def update_movie_session(session_id: int, show_time: str = None,
+                         movie_id: int = None,
+                         cinema_hall_id: int = None) -> MovieSession:
```
Nattalli, line 36: but your function returns nothing. `-> None` or add `return movie_session`

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
...
+        movie = Movie.objects.get(id=movie_id)
+        movie_session.movie = movie
+    if cinema_hall_id:
+        cinema_hall = CinemaHall.objects.get(id=cinema_hall_id)
+        movie_session.cinema_hall = cinema_hall
+    movie_session.save()
+
+
+def delete_movie_session_by_id(session_id: int) -> MovieSession:
+    if session_id:
```
Nattalli, line 51: unnecessary condition

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
...
+        movie_session = MovieSession.objects.filter(id=session_id).delete()
```
Nattalli, line 52: use `.get()` here
read the third item in checklist :

Nattalli (changes requested):

db/models.py
```diff
@@ -14,3 +14,34 @@ class Actor(models.Model):
...
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
+
+    def __str__(self) -> str:
+        return f"{self.title}"
```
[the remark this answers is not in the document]
alexzubkoff: Thanks! What a usefull link.

services/movie_session.py
```diff
@@ -6,48 +6,45 @@
...
-        movie_session.save()
-        return movie_session
+    cinema_hall_id = CinemaHall.objects.get(id=cinema_hall_id)
+    movie_id = Movie.objects.get(id=movie_id)
+    movie_session = MovieSession.objects.create(
+        show_time=movie_show_time,
+        cinema_hall=cinema_hall_id,
+        movie=movie_id
+    )
+    return movie_session
```
Nattalli, line 16: btw, see, it is the same code:

```python
movie_session = MovieSession.objects.create(
        show_time=movie_show_time,
        cinema_hall_id=cinema_hall_id,
        movie_id=movie_id
    )
```

Nattalli: so you don't need to overload your db with 
```python
cinema_hall_id = CinemaHall.objects.get(id=cinema_hall_id)
movie_id = Movie.objects.get(id=movie_id)
```

Nattalli: as foreignKey saves not a full object but a primary key only. By default primary key is an ID. In `create_movie_session` you take `movie_id` - the id of `Movie` class object. So 
Bad example of ForeignKey saving:
```python
movie_session = MovieSession.objects.create(
        show_time=movie_show_time,
        cinema_hall=CinemaHall.objects.get(id=cinema_hall_id),
        movie=Movie.objects.get(id=movie_id)
    )
```

Good example:
```python
movie_session = MovieSession.objects.create(
        show_time=movie_show_time,
        cinema_hall_id=cinema_hall_id,
        movie_id=movie_id
    )
```

Nattalli: as in the second example, you make 1 request to DB, at the first - 3 requests. Feel the difference now?

services/movie_session.py
```diff
@@ -6,48 +6,45 @@
...
 
 
-def get_movies_sessions(session_date: str = None) -> QuerySet | MovieSession:
+def get_movies_sessions(session_date: str = None) -> QuerySet:
     query_set = MovieSession.objects.all()
     if session_date:
-        query_set = query_set.filter(show_time__date=session_date)
+        query_set = MovieSession.objects.filter(show_time__date=session_date)
```
Nattalli, line 22: your previous solution was better

services/movie_session.py
```diff
@@ -6,48 +6,45 @@
...
                          cinema_hall_id: int = None) -> MovieSession:
-    if session_id:
-        movie_session = MovieSession.objects.get(id=session_id)
+    movie_session = MovieSession.objects.get(id=session_id)
     if show_time:
         movie_session.show_time = show_time
+
     if movie_id:
         movie = Movie.objects.get(id=movie_id)
         movie_session.movie = movie
```
Nattalli, line 39: and here:
```python
  if movie_id:
```

services/movie_session.py
```diff
@@ -6,48 +6,45 @@
...
+
     if cinema_hall_id:
-        cinema_hall = CinemaHall.objects.get(id=cinema_hall_id)
-        movie_session.cinema_hall = cinema_hall
+        cinema_hall_id = CinemaHall.objects.get(id=cinema_hall_id)
+        movie_session.cinema_hall = cinema_hall_id
```
Nattalli, line 43: optimize here too

Nattalli (changes requested):

services/movie_session.py
alexzubkoff: Thanks for the intelligible explanation! I probably did not quite correctly understand the conditions of the task, I thought that any data could come from outside, and before climbing into the database, they need to be checked whether there is such an id or not.

Nattalli: I understand that the way you did it looks much more logical. But in this case, other skills and understanding need to be developed here. + in Django, many things are already done so that errors are handled immediately

Nattalli (approved):
Good job!

PR #146, opened 2022-10-24
Solution

pavlo-kapitanets:


db/models.py
```diff
@@ -14,3 +14,36 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField(blank=True)
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
```
matechris, line 23: add related names for actors and genres

db/models.py
```diff
@@ -14,3 +14,36 @@ class Actor(models.Model):
...
+    def capacity(self) -> int:
+        return self.rows * self.seats_in_row
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall,
+                                    on_delete=models.SET_NULL,
+                                    null=True)
+    movie = models.ForeignKey(Movie, on_delete=models.SET_NULL, null=True)
```
matechris, line 46: add related names for cinema_hall and movie

services/movie.py
```diff
@@ -0,0 +1,37 @@
...
+               actors_ids: list[int] = None) -> QuerySet:
+    queryset = Movie.objects.all()
+    if genres_ids:
+        queryset = queryset.filter(genres__id__in=genres_ids)
+    if actors_ids:
+        queryset = queryset.filter(actors__id__in=actors_ids)
+    if actors_ids and genres_ids:
+        queryset = queryset.filter(
+            genres__id__in=genres_ids).filter(
+            actors__id__in=actors_ids)
```
matechris, line 16: I don't think it's necessary to have an additional check for both actors_ids and genres_ids

services/movie_session.py
```diff
@@ -0,0 +1,49 @@
...
+        cinema_hall=CinemaHall.objects.get(id=cinema_hall_id)
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
+    movie_sessions = MovieSession.objects.all()
+    if session_date:
+        movie_sessions = MovieSession.objects.filter(
+            show_time__date=session_date
+        )
```
matechris, line 21: you can use created variable in this if-statement:
`movie_sessions = movie_sessions.filter(...)`

matechris (changes requested):

db/models.py
pavlo-kapitanets: Thanks for your review! I've added all changes to my code.

matechris (approved):
Well done!

MasterpieceElbow (approved):
GJ!

PR #150, opened 2022-10-25
Solution

pazyn404:


services/movie.py
```diff
@@ -0,0 +1,41 @@
...
+
+def create_movie(movie_title: str,
+                 movie_description: str,
+                 genres_ids: list[int] = None,
+                 actors_ids: list[int] = None) -> Movie:
+    new_movie = Movie(
+        title=movie_title,
+        description=movie_description
+    )
+    new_movie.save()
```
Nattalli, line 34: Movie.objects.create()

services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+                         movie_id: int,
+                         cinema_hall_id: int) -> MovieSession:
+    new_movie_session = MovieSession(
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
+
+    new_movie_session.save()
+
```
Nattalli, line 18: MovieSession.objects.create()

services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+    return new_movie_session
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
+    if not session_date:
+        return MovieSession.objects.all()
+
+    desired_date = datetime.strptime(session_date, "%Y-%m-%d").date()
+
+    return MovieSession.objects.filter(show_time__date=desired_date)
```
Nattalli, line 28: `desired_date` is unnecessary
you can just filter with `show_time__date=session_date`

Nattalli (changes requested):

Nattalli (changes requested):
did you add only migrations? what about changes?

db/models.py
```diff
@@ -14,3 +14,34 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="actors")
```
Y-Havryliv, line 22: you still have wrong `related_name`

db/models.py
```diff
@@ -14,3 +14,34 @@ class Actor(models.Model):
...
+    genres = models.ManyToManyField(Genre, related_name="genres")
```
Y-Havryliv, line 23: same here

Y-Havryliv (changes requested):

db/models.py
pazyn404: Hi! What names threre should be? I tested it and got queryset with actors and genres respectively.

pazyn404: fix

Nattalli (approved):

PR #151, opened 2022-10-25
Develop

Ilyakson:


services/movie_session.py
```diff
@@ -0,0 +1,52 @@
...
+        queryset = queryset.filter(show_time__date=session_date)
+
+    return queryset
+
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
```
Vasyl-Poremchuk, line 31: And I think this function should also `return` a value.

Vasyl-Poremchuk (changes requested):

Vasyl-Poremchuk (approved):

db/models.py
```diff
@@ -14,3 +14,42 @@ class Actor(models.Model):
...
+    cinema_hall = models.ForeignKey(
+        CinemaHall,
+        related_name="cinema_hall_name",
+        on_delete=models.CASCADE
+    )
+    movie = models.ForeignKey(
+        Movie,
+        related_name="movie_name",
+        on_delete=models.CASCADE
+    )
```
TPolina, line 52: related_name is incorrect

TPolina (approved):
Well done!

PR #154, opened 2022-11-02
have 62 warnings its normal? tests passed

maxitoska:


services/movie.py
```diff
@@ -0,0 +1,40 @@
+from typing import List
+
+from db.models import Movie
+
+
+def get_movies(genres_ids: List[int] = None,
+               actors_ids: List[int] = None
+               ) -> Movie:
+    if genres_ids is None and actors_ids is None:
```
IvanRamyk, line 9: create a queryset with all entries at the begging and apply different filters depending on the values of the args. 
this approach is much easier to read

IvanRamyk (approved):
Well done!

PR #170, opened 2022-11-26
solution

KolyaZakharov:


services/movie.py
```diff
@@ -0,0 +1,32 @@
+from typing import List
+from db.models import Movie
+from django.db.models.query import QuerySet
+
+
+def get_movies(genres_ids: list = None,
+               actors_ids: List[int] = None) -> QuerySet:
+    queryset = Movie.objects.all()
+    if genres_ids is not None:
```
OleksandrPozden, line 9:
```suggestion
    if not genres_ids :

```

OleksandrPozden: the same in another case

services/movie_session.py
```diff
@@ -0,0 +1,40 @@
...
+from django.db.models.query import QuerySet
+
+
+def create_movie_session(movie_show_time: str,
+                         movie_id: int,
+                         cinema_hall_id: int) -> None:
+    MovieSession.objects.create(show_time=movie_show_time,
+                                cinema_hall_id=cinema_hall_id,
+                                movie_id=movie_id)
+
```
OleksandrPozden, line 11: change indentation to make it more readable

OleksandrPozden (changes requested):

services/movie.py
KolyaZakharov: in case  "if not genres_ids :"  gives an error  TypeError: 'NoneType' object is not iterable

OleksandrPozden (approved):

PR #178, opened 2022-12-13
solution

taras-zelinskyi:


services/movie_session.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def update_movie_session(session_id: int,
+                         show_time: str = None,
+                         movie_id: int = None,
+                         cinema_hall_id: int = None) -> None:
+    movie_session = MovieSession.objects.get(id=session_id)
+    if show_time is not None:
+        movie_session.show_time = show_time
+        movie_session.save()
```
IvanRamyk, line 35: perform `save` one time after all checks

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #184, opened 2022-12-24
Solution

yessloss:


services/cinema_hall.py
```diff
@@ -0,0 +1,13 @@
...
+def get_cinema_halls() -> None:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(hall_name: str,
+                       hall_rows: int,
+                       hall_seats_in_row: int) -> None:
+    return CinemaHall.objects.create(name=hall_name,
+                                     rows=hall_rows,
+                                     seats_in_row=hall_seats_in_row)
```
OleksandrPozden, line 13: fix indentation
or as the optional solution, you can have this
```suggestion
def create_cinema_hall(*args, **kwargs) -> None:
    return CinemaHall.objects.create(*args,**kwargs)
```
Also, fix here type annotation. It should be something else not None

services/movie.py
```diff
@@ -0,0 +1,30 @@
+from db.models import Movie
+
+
+def get_movies(genres_ids: list[int] = None,
+               actors_ids: list[int] = None) -> None:
```
OleksandrPozden, line 5: here as well. fix type annotaion

services/movie.py
```diff
@@ -0,0 +1,30 @@
...
+    return all_movie
+
+
+def get_movie_by_id(movie_id: int) -> None:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(movie_title: str, movie_description: str,
+                 genres_ids: list[int] = None,
+                 actors_ids: list[int] = None) -> None:
```
OleksandrPozden, line 23: type annotation and indentation

services/movie_session.py
```diff
@@ -0,0 +1,37 @@
+from db.models import MovieSession
+
+
+def create_movie_session(movie_show_time: None,
+                         movie_id: int, cinema_hall_id: int) -> None:
+    return MovieSession.objects.create(show_time=movie_show_time,
+                                       cinema_hall_id=cinema_hall_id,
+                                       movie_id=movie_id)
```
OleksandrPozden, line 8: the same

services/movie_session.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def get_movies_sessions(session_date: str = None) -> None:
+    movie_session = MovieSession.objects.all()
+    if session_date:
+        return movie_session.filter(show_time__date=session_date)
+    return movie_session
```
OleksandrPozden, line 15: the same here and for others

OleksandrPozden (changes requested):

services/cinema_hall.py
yessloss: what happen with indentation?

andrewbelo (approved):

PR #187, opened 2023-01-05
Solution

Nexxel05:


services/movie.py
```diff
@@ -0,0 +1,48 @@
...
+
+    queryset = Movie.objects.all()
+
+    if genres_ids and actors_ids:
+        queryset = queryset.filter(
+            genres__id__in=genres_ids,
+            actors__id__in=actors_ids
+        )
+
+    else:
```
YuriiKindrat, line 18: Unnecessary part of code - the queryset will be filtered by genres and actors only using 19-23 lines

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #204, opened 2023-01-28
Solution2

Tania-Kharchuk:


services/cinema_hall.py
```diff
@@ -0,0 +1,18 @@
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> list[CinemaHall]:
```
olena-hul, line 4: It won't be list, it will be Queryset[CinemaHall[

services/movie.py
```diff
@@ -0,0 +1,38 @@
+from db.models import Movie
+
+
+def get_movies(genres_ids: list[int] = None,
+               actors_ids: list[int] = None
```
olena-hul, line 5: Add Optional to annotation, because you're setting None as default. And fix return annotation, it will be Queryset

services/movie_session.py
```diff
@@ -0,0 +1,49 @@
...
+                         ) -> MovieSession:
+    return MovieSession.objects.create(
+        show_time=movie_show_time,
+        cinema_hall_id=cinema_hall_id,
+        movie_id=movie_id
+    )
+
+
+def get_movies_sessions(session_date: str = None
+                        ) -> MovieSession | list[MovieSession]:
```
olena-hul, line 16:
```suggestion
def get_movies_sessions(
    session_date: str = None
) -> MovieSession | Queryset[MovieSession]:
```

services/movie.py
olena-hul: Fix all cases with `= None` in args, add Optional

services/movie_session.py
```diff
@@ -0,0 +1,49 @@
...
+def update_movie_session(session_id: int,
+                         show_time: str = None,
+                         movie_id: int = None,
+                         cinema_hall_id: int = None
+                         ) -> None:
+    updated_movie_session = MovieSession.objects.get(id=session_id)
+
+    if show_time:
+        updated_movie_session.show_time = show_time
+        updated_movie_session.save()
```
olena-hul, line 37: You can call .save() once in the end of the method

olena-hul (changes requested):

olena-hul (approved):

olena-hul:
Good job!

PR #205, opened 2023-01-31
Solution

anastasiia-tsurkan:


services/movie.py
```diff
@@ -0,0 +1,40 @@
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
```
ArsenPidhoretskyi, line 5: Optional[list[int]]
Also the same with actors_ids

ArsenPidhoretskyi: In create_movie the same problem

services/movie_session.py
```diff
@@ -0,0 +1,56 @@
...
+    new_movie_session = MovieSession.objects.create(
+        show_time=movie_show_time,
+        cinema_hall=cinema_hall,
+        movie=movie
+    )
+
+    return new_movie_session
+
+
+def get_movies_sessions(session_date: str = None) -> MovieSession:
```
ArsenPidhoretskyi, line 22: It returns queryset

services/movie_session.py
```diff
@@ -0,0 +1,56 @@
...
+        movie_id: int = None,
+        cinema_hall_id: int = None,
+) -> None:
+    movie_session_to_update = MovieSession.objects.get(
+        id=session_id
+    )
+
+    if show_time:
+        movie_session_to_update.show_time = show_time
+        movie_session_to_update.save()
```
ArsenPidhoretskyi, line 44: no need to save in each if
save after

ArsenPidhoretskyi (changes requested):

services/movie_session.py
anastasiia-tsurkan: In the way of using keyword arguments to the class save() method is needed

services/movie.py
anastasiia-tsurkan: Danyl shows  in videos that "optional" was writen as I wrote through None. Can you please share more info about Optional, please?

andrewbelo (changes requested):
Попередні зауваження не виправлені

services/movie.py
ArsenPidhoretskyi: https://stackoverflow.com/questions/51710037/how-should-i-use-the-optional-type-hint
Also screenshot
image

services/movie_session.py
ArsenPidhoretskyi: Yeah but call it at the end of the function
Because for now you have duplicates and after each change do a query in the database
lines 45,49, 53

ArsenPidhoretskyi (changes requested):

services/movie_session.py
ArsenPidhoretskyi: @anastasiia-tsurkan not fixed

ArsenPidhoretskyi (changes requested):
Last comment)

olena-hul:
Good job!

PR #206, opened 2023-02-02
Solution

YuriiVataschuk:


services/cinema_hall.py
```diff
@@ -0,0 +1,19 @@
+import init_django_orm  # noqa: F401
+
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> CinemaHall:
```
ArsenPidhoretskyi, line 6: It returns queryset not CinemaHall
Also, .all() can not return one object

services/movie.py
```diff
@@ -0,0 +1,39 @@
+import init_django_orm  # noqa: F401
+from django.db.models.query import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
```
ArsenPidhoretskyi, line 8: Optional[list[int]]
`genres_ids = None` it is already mean that it must accept optional
The same in all other places

ArsenPidhoretskyi (changes requested):

services/movie.py
YuriiVataschuk: > Optional[list[int]] `genres_ids = None` it is already mean that it must accept optional The same in all other places

Can you explain please, Do I have to replace genres_ids: list[int] = None with Optional[list[int]]?

Grayder0152: > > Optional[list[int]] `genres_ids = None` it is already mean that it must accept optional The same in all other places
> 
> Can you explain please, Do I have to replace genres_ids: list[int] = None with Optional[list[int]]?

Yes

Grayder0152 (approved):

PR #211, opened 2023-02-07
'Solution'

Oleksiy-Liubchenko:


services/movie.py
```diff
@@ -0,0 +1,44 @@
...
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
+) -> QuerySet:
+    queryset = Movie.objects.all()
+    if genres_ids and actors_ids:
+        return queryset.filter(
+            genres__id__in=genres_ids
+        ).filter(
+            actors__id__in=actors_ids
+        )
```
Bohdan-Salamakha, line 15: not queryset.filter(
            genres__id__in=genres_ids
        ).filter(
            actors__id__in=actors_ids
        )
        but queryset.filter(
            genres__id__in=genres_ids, actors__id__in=actors_ids
        )

services/movie.py
```diff
@@ -0,0 +1,44 @@
...
+        return queryset.filter(genres__id__in=genres_ids)
+
+    if actors_ids:
+        return queryset.filter(actors__id__in=actors_ids)
+
+    return queryset
+
+
+def get_movie_by_id(movie_id: int) -> object:
+    return Movie.objects.get(id=movie_id)
```
Bohdan-Salamakha, line 27: if object doesn't exist u will get an internal server error, if u don't handle it in the middleware u should use
get_object_or_404() func instead

services/movie_session.py
```diff
@@ -0,0 +1,48 @@
...
+
+def get_movies_sessions(session_date: datetime = None) -> QuerySet:
+    queryset = MovieSession.objects.all()
+    if session_date:
+        return queryset.filter(show_time__date=session_date)
+    return queryset
+
+
+def get_movie_session_by_id(movie_session_id: int) -> object:
+    return MovieSession.objects.get(id=movie_session_id)
```
Bohdan-Salamakha, line 26: if object doesn't exist u will get an internal server error, if u don't handle it in the middleware u should use
get_object_or_404() func instead

Bohdan-Salamakha (changes requested):
v1

services/movie.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def get_movie_by_id(movie_id: int) -> object:
+    return get_object_or_404(Movie, id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: list[int] = None,
```
olena-hul, line 33:
```suggestion
        genres_ids: Optional[list[int]] = None,

```

services/movie_session.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def get_movies_sessions(session_date: datetime = None) -> QuerySet:
+    queryset = MovieSession.objects.all()
+    if session_date:
+        return queryset.filter(show_time__date=session_date)
+    return queryset
+
+
+def get_movie_session_by_id(movie_session_id: int) -> object:
```
olena-hul, line 26: object is bad annotation

olena-hul (changes requested):

Bohdan-Salamakha (approved):

PR #217, opened 2023-02-15
Solution

leetwinoff:


services/movie_session.py
```diff
@@ -0,0 +1,52 @@
...
+        show_time=movie_show_time,
+        cinema_hall_id=cinema_hall_id,
+        movie_id=movie_id
+    )
+
+
+def get_movies_sessions(
+        session_date: str = None
+) -> MovieSession:
+    sessions = MovieSession.objects.all()
```
katevasilenko, line 20: You can get it directly on return

services/movie_session.py
```diff
@@ -0,0 +1,52 @@
...
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: datetime = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> MovieSession:
+    session = MovieSession.objects.get(id=session_id)
```
katevasilenko, line 38: Use get_movie_session_by_id() function

katevasilenko (changes requested):

services/movie.py
```diff
@@ -0,0 +1,35 @@
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
ArsenPidhoretskyi, line 6: Optional[list[int]] because you already have default parameter None

services/movie.py
```diff
@@ -0,0 +1,35 @@
...
+) -> Movie:
```
ArsenPidhoretskyi, line 7: It will return queryset

services/movie.py
ArsenPidhoretskyi: The same on other places

services/movie_session.py
```diff
@@ -0,0 +1,51 @@
...
+    return MovieSession.objects.create(
+        show_time=movie_show_time,
+        cinema_hall_id=cinema_hall_id,
+        movie_id=movie_id
+    )
+
+
+def get_movies_sessions(
+        session_date: str = None
+) -> MovieSession:
```
ArsenPidhoretskyi, line 19: It will return queryet

services/movie_session.py
```diff
@@ -0,0 +1,51 @@
...
+
+    if session_date is not None:
+        session_date = datetime.strptime(session_date, "%Y-%m-%d")
```
ArsenPidhoretskyi, line 22: Why we need this convertion?

services/movie_session.py
```diff
@@ -0,0 +1,51 @@
...
+    if cinema_hall_id is not None:
+        session.cinema_hall_id = cinema_hall_id
+    session.save()
+    return session
+
+
+def delete_movie_session_by_id(
+        session_id: int
+) -> None:
+    MovieSession.objects.filter(id=session_id).delete()
```
ArsenPidhoretskyi, line 51: Reuse get_movie_session_by_id(session_id)

ArsenPidhoretskyi (changes requested):

Vasyl-Smutok (approved):
Well Done 👍

PR #220, opened 2023-02-16
Solution

ArtMot99:


services/movie.py
```diff
@@ -0,0 +1,49 @@
...
+    # If only actors_ids is provided
+    if genres_ids is None and actors_ids:
+        queryset = queryset.filter(actors__id__in=actors_ids)
+
+    # If both genres_ids and actors_ids are provided
+    if genres_ids and actors_ids:
+        queryset = queryset.filter(
+            genres__id__in=genres_ids,
+            actors__id__in=actors_ids
+        )
```
ArsenPidhoretskyi, line 25: You can simplify this and write only with 2 conditions:
1. genres_ids is not None
2. actors_ids is not None

services/cinema_hall.py
```diff
@@ -0,0 +1,18 @@
+from db.models import CinemaHall
+from typing import Any
+
+
+def get_cinema_halls() -> Any:
```
ArsenPidhoretskyi, line 5: No, specify the correct type. It does not return Any it return QuerySet

services/movie.py
```diff
@@ -0,0 +1,39 @@
+from db.models import Movie
+from typing import Any, Optional
+
+
+def get_movies(
+        genres_ids: Optional[list[int]] = None,
+        actors_ids: Optional[list[int]] = None
+) -> Any:
```
ArsenPidhoretskyi, line 8: Any is not accepted if you can provide more detailed type

ArsenPidhoretskyi: Fix in all places

ArsenPidhoretskyi (changes requested):

Bohdan-Salamakha (changes requested):
don't work on the master branch - it's a bad practice
https://github.com/mate-academy/py-task-guideline/blob/main/README.md#:~:text=git%20checkout%20%2Db%20develop

PR #223, opened 2023-02-20
Solution

ViacheslavYasynetskyi:


db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
```
vasylhnatiuk, line 22: you must add related_name for actors and genres

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
...
+        return self.name
+
+    @property
+    def capacity(self) -> int:
+        return self.rows * self.seats_in_row
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall, on_delete=models.CASCADE)
```
vasylhnatiuk, line 44: the same here for cinema_hall and movie

services/movie.py
```diff
@@ -0,0 +1,39 @@
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list = None,
```
vasylhnatiuk, line 5: must be list[int]

services/movie.py
```diff
@@ -0,0 +1,39 @@
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list = None,
+        actors_ids: list = None
```
vasylhnatiuk, line 6: same here

vasylhnatiuk (changes requested):

db/models.py
ArsenPidhoretskyi: @ViacheslavYasynetskyi not fixed

db/models.py
ArsenPidhoretskyi: @ViacheslavYasynetskyi not fixed

services/movie.py
ArsenPidhoretskyi, line 5: It is Optional[list[int]] it accepts None by default
Do not forgot add inner annotations for list

ArsenPidhoretskyi: Fix in other places

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Good job

PR #225, opened 2023-02-22
Solution

Sebshe:


services/movie.py
```diff
@@ -0,0 +1,39 @@
+from db.models import Movie
+from django.db.models import QuerySet
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
ArsenPidhoretskyi, line 7: They are Optional parameters because as default they accept None

ArsenPidhoretskyi: Fix the same in other places

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (commented):
Well done!)

ArsenPidhoretskyi (approved):

PR #227, opened 2023-02-24
Solution

andrii-skryl:


services/movie.py
```diff
@@ -0,0 +1,44 @@
+from django.db.models import QuerySet
+
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
ArsenPidhoretskyi, line 9: Accepts Optional[list[int]]
If default parameter is None the typing must include optional
Change in other places

services/movie.py
```diff
@@ -0,0 +1,44 @@
...
+    movies = Movie.objects.all()
+    if genres_ids is not None and actors_ids is not None:
+        return movies.filter(
+            genres__id__in=genres_ids,
+            actors__id__in=actors_ids
+        ).distinct()
+    if genres_ids is not None and actors_ids is None:
+        return movies.filter(genres__id__in=genres_ids).distinct()
+    if genres_ids is None and actors_ids is not None:
+        return movies.filter(actors__id__in=actors_ids).distinct()
```
ArsenPidhoretskyi, line 21: You have two cases here no need to hard ifs:
1. if genres_ids is not None
2. if actors_ids is not None

ArsenPidhoretskyi (changes requested):

Bohdan-Salamakha (approved):

Abnormaltype (approved):

PR #228, opened 2023-02-26
Develop

frameq:


services/movie.py
```diff
@@ -0,0 +1,34 @@
+from db.models import Movie
+from typing import List
+from django.db.models import QuerySet
+
+
+def get_movies(
+    genres_ids: List[int] = None, actors_ids: List[int] = None
```
Grayder0152, line 7: Use Optional

db/models.py
```diff
@@ -14,3 +14,38 @@ class Actor(models.Model):
...
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="actors")
+    genres = models.ManyToManyField(Genre, related_name="genres")
+
+    def __str__(self) -> str:
+        return f"{self.title}"
```
katevasilenko, line 26: Actually you don't need f string here

db/models.py
```diff
@@ -14,3 +14,38 @@ class Actor(models.Model):
...
+    name = models.CharField(max_length=255)
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
+
+    @property
+    def capacity(self) -> int:
+        return self.rows * self.seats_in_row
+
+    def __str__(self) -> str:
+        return f"{self.name}"
```
katevasilenko, line 39: The same? Remove f string

services/movie.py
```diff
@@ -0,0 +1,35 @@
+from db.models import Movie
+from django.db.models import QuerySet
+from typing import List, Optional
```
katevasilenko, line 3: Please sort imports: built-in is first, third-party next, you own imports is the last. Blank line between each block

services/movie_session.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def update_movie_session(
+    session_id: int,
+    show_time: Optional[datetime] = None,
+    movie_id: Optional[int] = None,
+    cinema_hall_id: Optional[int] = None,
+) -> None:
+    query = {
+        f"{key}": value for key, value in locals().items() if value is not None
```
katevasilenko, line 40: I don’t get this. Looks too complicated for this task. Don’t use locals()

katevasilenko (changes requested):

PR #230, opened 2023-02-27
Solution

VItalik1292:


services/movie.py
```diff
@@ -0,0 +1,40 @@
+from db.models import Movie
+from django.db.models import QuerySet
+
+
+def get_movies(
+        genres_ids: list[int] = None,
```
ArsenPidhoretskyi, line 6: It also accepts None then it is mean Optional[list[int]] 
Check also in other places

ArsenPidhoretskyi (changes requested):

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
...
+    return MovieSession.objects.all()
+
+
+def get_movie_session_by_id(movie_session_by_id: int) -> QuerySet:
+    return MovieSession.objects.get(id=movie_session_by_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: datetime = None,
```
ArsenPidhoretskyi, line 32: Optional[datetime]

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
...
+        movie_id: Optional[int] = None,
+        cinema_hall_id: Optional[int] = None,
+) -> QuerySet:
+    queryset = MovieSession.objects.all()
+    if show_time is not None:
+        queryset = MovieSession.objects.filter(
```
ArsenPidhoretskyi, line 38: no need to again use MovieSession.objects.filter
use instead queryset = queryset.filter(

ArsenPidhoretskyi: The same in other places

ArsenPidhoretskyi (changes requested):

db/models.py
```diff
@@ -14,3 +14,36 @@ class Actor(models.Model):
...
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall, on_delete=models.CASCADE)
+    movie = models.ForeignKey(Movie, on_delete=models.CASCADE)
+
+    def __str__(self) -> str:
+        return f"{self.movie.title} " \
+               f"{self.show_time.strftime('%Y-%m-%d %H:%M:%S')}"
```
Vasyl-Smutok, line 49: better to usr `()` then `\`
```suggestion
        return (
            f"{self.movie.title} "
            f"{self.show_time.strftime('%Y-%m-%d %H:%M:%S')}"
        )
```

Vasyl-Smutok (approved):
Good Job 👍

PR #231, opened 2023-02-28
Solution

bihunva:


services/movie_session.py
```diff
@@ -0,0 +1,49 @@
...
+        cinema_hall_id: int
+) -> None:
+    MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> MovieSession | QuerySet:
```
YuriiKindrat, line 20: are u sure that this method can return MovieSession obj ?

YuriiKindrat (changes requested):

services/movie_session.py
bihunva: fixed

wQuelS (approved):
Good

PR #237, opened 2023-03-03
Solution

SlavikSherbak:


services/movie.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: Optional[list[int]] = None,
```
vasylhnatiuk, line 30: would be better to use `genres_ids: List[int]`. here and other places

services/movie_session.py
```diff
@@ -0,0 +1,51 @@
...
+    return query_set
+
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: Optional[str] = None,
```
vasylhnatiuk, line 35: it must be `show_time: datetime = None,`

services/movie_session.py
```diff
@@ -0,0 +1,51 @@
...
+        movie_id: Optional[int] = None,
```
vasylhnatiuk, line 36: remove `Optional` here and other places

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):
Well done!

PR #242, opened 2023-03-07
Solution

avkpol:


services/movie.py
```diff
@@ -0,0 +1,45 @@
+from typing import Optional
+
+from django.db.models import QuerySet
+
+from db.models import Movie, Genre, Actor
+
+
+def get_movies(
+        genres_ids: Optional[list] = None,
```
ArsenPidhoretskyi, line 9:
```suggestion
        genres_ids: Optional[list[int]] = None,
```

ArsenPidhoretskyi: In other places the same

services/movie_session.py
```diff
@@ -0,0 +1,58 @@
...
+    return movie_sessions
+
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: str = None,
```
ArsenPidhoretskyi, line 40: Optional[str]
The same with other types

ArsenPidhoretskyi (changes requested):

db/models.py
```diff
@@ -14,3 +14,36 @@ class Actor(models.Model):
...
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall, on_delete=models.CASCADE)
+    movie = models.ForeignKey(Movie, on_delete=models.CASCADE)
+
+    def __str__(self) -> str:
+        return f"{self.movie.title} " \
+               f"{self.show_time.strftime('%Y-%m-%d %H:%M:%S')}"
```
Vasyl-Smutok, line 49: better to use `()` then `\`

Vasyl-Smutok (changes requested):

services/movie_session.py
```diff
@@ -0,0 +1,61 @@
...
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: Optional[str] = None,
+        movie_id: Optional[int] = None,
+        cinema_hall_id: Optional[int] = None) -> MovieSession:
+    movie_session = MovieSession.objects.get(id=session_id)
```
katevasilenko, line 46: You also can use get_movie_session_by_id here

katevasilenko (approved):
Looks good to me, just one advice

services/movie_session.py
avkpol: agree

PR #244, opened 2023-03-08
Solution

AlimFreight:


services/movie.py
```diff
@@ -0,0 +1,42 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
ArsenPidhoretskyi, line 8: It accepts Optional[list[int]] because by default is None
Change in other places(with all types)

services/movie_session.py
```diff
@@ -0,0 +1,65 @@
...
+    )
+    return movie_session
+
+
+def get_movies_sessions(
+        session_date: dt.datetime = None
+) -> QuerySet:
+    movie_session = MovieSession.objects.all()
+    if session_date:
+        movie_session = MovieSession.objects.filter(
```
ArsenPidhoretskyi, line 25: No need to write again MoviSession.objects. better to do as 
```suggestion
        movie_session = movie_session.filter(
```

services/movie_session.py
```diff
@@ -0,0 +1,65 @@
...
+            show_time__date=session_date
+        )
+    return movie_session
+
+
+def get_movie_session_by_id(session_id: int) -> QuerySet:
```
ArsenPidhoretskyi, line 31: It returns object(MovieSession) not QuerySet

services/movie_session.py
```diff
@@ -0,0 +1,65 @@
...
+    session = MovieSession.objects.filter(
+        id=session_id,
+    )
+    if cinema_hall_id:
+        session.update(cinema_hall_id=cinema_hall_id)
+    if movie_id:
+        session.update(movie_id=movie_id)
+    if show_time:
+        session.update(show_time=show_time)
+
```
ArsenPidhoretskyi, line 56: Need to do this on one query, because for now it works with 3 queries

services/movie_session.py
```diff
@@ -0,0 +1,65 @@
...
+        movie_id: int = None,
+        cinema_hall_id: int = None,
+) -> QuerySet:
+    if not any([cinema_hall_id, show_time, movie_id]):
+        raise ValueError(
+            "You won't be able to update without"
+            " providing at least one argument"
+        )
+
+    session = MovieSession.objects.filter(
```
ArsenPidhoretskyi, line 47: You can use .get

ArsenPidhoretskyi (changes requested):

services/movie_session.py
Grayder0152: pls, resolve comment

Grayder0152 (changes requested):

Grayder0152 (approved):

PR #250, opened 2023-03-08
Develop

Mazurinka29:


services/movie_session.py
```diff
@@ -0,0 +1,65 @@
...
+    return movie_session
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: datetime.datetime = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> None:
+    get_movie = MovieSession.objects.get(id=session_id)
```
olena-hul, line 46:
```suggestion
    movie = MovieSession.objects.get(id=session_id)
```

olena-hul (changes requested):

services/movie_session.py
```diff
@@ -0,0 +1,63 @@
...
+
+def get_movies_sessions(session_date: str = None) -> QuerySet[MovieSession]:
+    all_movies = MovieSession.objects.all()
+
+    if session_date is not None:
+        filtered_movies = MovieSession.objects.filter(
+            show_time__date=session_date
+        )
+        return filtered_movies
+    return all_movies
```
ArsenPidhoretskyi, line 32: This is not good better to do as:
```suggestion
    if session_date is not None:
        all_movies = all_movies.filter(
            show_time__date=session_date
        )
    return all_movies
```

ArsenPidhoretskyi: Because if you will here so much filters it will be so strange

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #251, opened 2023-03-09
solution

DavidGlusch:


services/movie.py
```diff
@@ -0,0 +1,39 @@
+from django.db.models import QuerySet
+
+from db.models import Genre, Actor, Movie
+
+
+def get_movies(
+        genres_ids: list[Genre] = None,
+        actors_ids: list[Actor] = None
```
amber-marichi, line 8: maybe i dont get it but imo this should not be hinted as "list of Genre instances", since inside the function you are clearly comparing them to ids but not instances....

services/cinema_hall.py
```diff
@@ -0,0 +1,19 @@
...
+
+def get_cinema_halls() -> QuerySet:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(
+        hall_name: str,
+        hall_rows: int,
+        hall_seats_in_row: int
+) -> None:
```
amber-marichi, line 14: but you are returning created object, right? please check the hinting

services/movie_session.py
```diff
@@ -0,0 +1,49 @@
+from django.db.models import QuerySet
+
+from db.models import MovieSession
+
+
+def create_movie_session(
+        movie_show_time: str,
+        movie_id: int,
+        cinema_hall_id: int,
+) -> None:
```
amber-marichi, line 10: here too, please check return hint

amber-marichi (approved):
good code, few questions on type hinting tho

db/models.py
```diff
@@ -14,3 +14,39 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
```
wQuelS, line 23: add `related_name`. It is common best practice

db/models.py
```diff
@@ -14,3 +14,39 @@ class Actor(models.Model):
...
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(
+        CinemaHall,
+        on_delete=models.CASCADE,
+
+    )
+    movie = models.ForeignKey(Movie, on_delete=models.SET_NULL, null=True)
```
wQuelS, line 49: why did u use `SET_NULL` here ?

services/movie.py
wQuelS, line 8: yeah, this should be list of `int`s

services/cinema_hall.py
wQuelS, line 14: but you do return something in this function. Also, you don't actually need that `return`. And if you get rid of it, the return type hint can remain as `None`

services/movie_session.py
wQuelS, line 10: same

wQuelS (changes requested):

db/models.py
```diff
@@ -14,3 +14,39 @@ class Actor(models.Model):
...
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(
+        CinemaHall,
+        on_delete=models.CASCADE,
+
+    )
+    movie = models.ForeignKey(Movie, on_delete=models.CASCADE, null=True)
```
Serhii-Leonenko, line 49: I think that movie is an obligatory field in this case, it can't be null

services/movie_session.py
```diff
@@ -0,0 +1,49 @@
...
+
+    if show_time:
+        movie_session.update(show_time=show_time)
+
+    if movie_id:
+        movie_session.update(movie_id=movie_id)
+
+    if cinema_hall_id:
+        movie_session.update(cinema_hall_id=cinema_hall_id)
+
```
Serhii-Leonenko, line 46: you can use .get() method to get the object, and then change the attributes of the object and then save(). So, in the result, you remove redundant queries

Serhii-Leonenko (changes requested):

services/movie_session.py
```diff
@@ -0,0 +1,51 @@
...
+        cinema_hall_id=cinema_hall_id,
+        movie_id=movie_id
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
+    if session_date:
+        return MovieSession.objects.filter(show_time__date=session_date)
+    return MovieSession.objects.all()
+
```
oksana-feshchenko, line 23: just a suggestion: use one return here.

oksana-feshchenko (approved):

YuriiKindrat (approved):

PR #252, opened 2023-03-09
solution

amber-marichi:


DmytryTong (approved):
Good job)

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="movies")
+    genres = models.ManyToManyField(Genre, related_name="genres")
```
YuriiKindrat, line 23: are u sure about related name ?
image

services/cinema_hall.py
```diff
@@ -0,0 +1,20 @@
...
+    hall_name: str,
+    hall_rows: int,
+    hall_seats_in_row: int
+) -> CinemaHall:
+    new_cinemahall = CinemaHall.objects.create(
+        name=hall_name,
+        rows=hall_rows,
+        seats_in_row=hall_seats_in_row
+    )
+    return new_cinemahall
```
YuriiKindrat, line 20: no need to create varaible here

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+    movie_show_time: datetime | str,
+    movie_id: int,
+    cinema_hall_id: int
+) -> MovieSession:
+    new_moviesession = MovieSession.objects.create(
+        show_time=movie_show_time,
+        cinema_hall_id=cinema_hall_id,
+        movie_id=movie_id
+    )
+    return new_moviesession
```
YuriiKindrat, line 17: same here

YuriiKindrat (changes requested):

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="actors")
+    genres = models.ManyToManyField(Genre, related_name="genres")
```
wQuelS, line 23: I guess both of `related_names` should be `movies`

wQuelS (changes requested):

db/models.py
amber-marichi: thanks, i checked out info on related names. seems clearer now

YuriiKindrat (approved):

PR #253, opened 2023-03-09
Develop

ochernyi:


services/movie.py
```diff
@@ -0,0 +1,41 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
```
ArsenPidhoretskyi, line 7: it is Optional[list[int]] because accepts default as None
Fix in other places(with all types)

services/movie.py
```diff
@@ -0,0 +1,41 @@
...
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: list = None,
+        actors_ids: list = None
```
ArsenPidhoretskyi, line 29: list of what?

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #255, opened 2023-03-09
Develop DonchykD

DmytryTong:


services/movie_session.py
```diff
@@ -0,0 +1,57 @@
+from datetime import datetime
+
+from db.models import MovieSession
+
+
+def create_movie_session(movie_show_time: datetime, movie_id: int, cinema_hall_id: int):
```
Morty67: Where are type annotations for returned value?

services/movie_session.py
```diff
@@ -0,0 +1,57 @@
...
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
+
+    return new_movie_session
+
+
+def get_movies_sessions(session_date: str = None):
+    queryset = MovieSession.objects.all()
```
Morty67: Where are type annotations for returned value?

Morty67 (changes requested):

services/movie_session.py
DmytryTong: Check lattest commit, please.

services/movie_session.py
DmytryTong: Check lattest commit, please.

Morty67: > Check lattest commit, please.

sry, my bad

db/models.py
```diff
@@ -14,3 +15,36 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
```
wQuelS, line 24: Add `related_name` as it is a good practice overall

services/cinema_hall.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def get_cinema_halls() -> QuerySet:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(hall_name: str,
+                       hall_rows: int,
+                       hall_seats_in_row: int) -> CinemaHall:
+    new_cinema_hall = CinemaHall.objects.create(
```
wQuelS, line 13: no need to use additional variable here.

services/movie_session.py
```diff
@@ -0,0 +1,61 @@
...
+
+from django.db.models import QuerySet
+
+from db.models import MovieSession
+
+
+def create_movie_session(movie_show_time: datetime,
+                         movie_id: int,
+                         cinema_hall_id: int) -> MovieSession:
+    new_movie_session = MovieSession.objects.create(
```
wQuelS, line 11: same about additional variable

wQuelS (changes requested):

db/models.py
DmytryTong: Done.

services/cinema_hall.py
DmytryTong: Fixed.

services/movie_session.py
```diff
@@ -0,0 +1,59 @@
...
+def get_movies_sessions(session_date: str = None) -> QuerySet:
+    queryset = MovieSession.objects.all()
+
+    if session_date:
+        year = session_date.split("-")[0]
+        month = session_date.split("-")[1]
+        day = session_date.split("-")[2]
+        queryset = queryset.filter(show_time__year=year,
+                                   show_time__month=month,
+                                   show_time__day=day,)
```
Serhii-Leonenko, line 27: try to use filter by `show_time__date` , you can remove these splits

services/movie_session.py
```diff
@@ -0,0 +1,59 @@
...
+
+    if movie_id:
+        queryset.update(
+            movie_id=movie_id
+        )
+
+    if cinema_hall_id:
+        queryset.update(
+            cinema_hall_id=cinema_hall_id
+        )
```
Serhii-Leonenko, line 55: you can use .get() method to get the object, and then change the attributes of the object and then save(). So, in the result, you remove redundant queries

Serhii-Leonenko (changes requested):

services/movie_session.py
DmytryTong: Done.

services/movie_session.py
DmytryTong: Done.

db/models.py
```diff
@@ -14,3 +15,36 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="actors")
+    genres = models.ManyToManyField(Genre, related_name="genres")
```
YuriiKindrat, line 24: are u sure about related names here? check checklist

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
...
+
+    if movie_id:
+        movie_session = MovieSession.objects.get(id=session_id)
+        movie_session.movie_id = movie_id
+        movie_session.save()
+
+    if cinema_hall_id:
+        movie_session = MovieSession.objects.get(id=session_id)
+        movie_session.cinema_hall_id = cinema_hall_id
+        movie_session.save()
```
YuriiKindrat, line 49: why do u need to make 3 query to a db ? why don't u use variable ?

YuriiKindrat (changes requested):

services/movie_session.py
DmytryTong: I did something like that already, but I was rejected. I'll change it back, but with a current .save() functions.

db/models.py
DmytryTong: Fixed.

services/movie_session.py
```diff
@@ -0,0 +1,51 @@
...
+        movie_session.show_time = show_time
+        movie_session.save()
+
+    if movie_id:
+        movie_session.movie_id = movie_id
+        movie_session.save()
+
+    if cinema_hall_id:
+        movie_session.cinema_hall_id = cinema_hall_id
+        movie_session.save()
```
YuriiKindrat, line 47: u can use save method only once here

YuriiKindrat (changes requested):

services/movie_session.py
DmytryTong: Fixed.

YuriiKindrat (approved):

PR #256, opened 2023-03-09
Solution

kuznetsb:


MikhailLyvak (approved):

services/movie_session.py
```diff
@@ -0,0 +1,47 @@
+from db.models import MovieSession
+
+from django.db.models import QuerySet
+from datetime import datetime
```
Serhii-Leonenko, line 4: fix imports according to PEP8

services/movie_session.py
```diff
@@ -0,0 +1,47 @@
...
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
+    query = MovieSession.objects.all()
+    if session_date:
+        date = datetime.strptime(session_date, "%Y-%m-%d")
```
Serhii-Leonenko, line 22: Will it work without this `date = datetime.strptime(session_date, "%Y-%m-%d")`?)

services/movie_session.py
```diff
@@ -0,0 +1,47 @@
...
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> None:
+    movie_session = MovieSession.objects.filter(id=session_id)
+    if show_time:
+        movie_session.update(show_time=show_time)
+    if movie_id:
+        movie_session.update(movie_id=movie_id)
+    if cinema_hall_id:
+        movie_session.update(cinema_hall_id=cinema_hall_id)
```
Serhii-Leonenko, line 43: you can use .get() method to get the object, and then change the attributes of the object and then save(). So, in the result, you remove redundant queries

Serhii-Leonenko (changes requested):

Danil228 (approved):

YuriiKindrat (approved):

PR #257, opened 2023-03-10
Develop

ilushasm:


services/movie.py
```diff
@@ -0,0 +1,40 @@
+from db.models import Movie
+
+
+def get_movies(
+    genres_ids: list[int] = None,
+    actors_ids: list[int] = None
```
ArsenPidhoretskyi, line 6: If None as default then it accepts Optional[
The same in other places

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):
Good job!

PR #260, opened 2023-03-10
Solution

MikhailLyvak:


services/movie_session.py
```diff
@@ -0,0 +1,51 @@
...
+        session_date: datetime = None
+) -> QuerySet:
+    if session_date:
+        return MovieSession.objects.filter(show_time__date=session_date)
+    return MovieSession.objects.all()
+
+
+def get_movie_session_by_id(
+        movie_session_id: int
+) -> QuerySet:
```
ostapT, line 30: when you get one movie session - return type of annotation should be class instance, not queryset

ostapT (changes requested):

PR #263, opened 2023-03-10
Develop

ostapT:


services/movie_session.py
```diff
@@ -0,0 +1,56 @@
...
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+    session_id: int,
+    show_time: str = None,
+    movie_id: int = None,
+    cinema_hall_id: int = None,
+) -> None:
+    movie_session = MovieSession.objects.all()
```
ZhAlexR, line 41: Why are you trying to get all sessions and then filtering them by 'session_id' to have a single one? It is simpler to use MovieSession.objects.get(id=session_id) - you will have an object instantly instead of filtering a QuerySet.

services/movie_session.py
```diff
@@ -0,0 +1,56 @@
...
+
+    if movie_id is not None:
+        movie_session.update(movie=movie_id)
+
+    if cinema_hall_id is not None:
+        movie_session.update(cinema_hall=cinema_hall_id)
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    movie_session = MovieSession.objects.all()
```
ZhAlexR, line 55: The same as above

ZhAlexR (approved):

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="actors")
+    genres = models.ManyToManyField(Genre, related_name="genres")
```
YuriiKindrat, line 23: check ur checklist

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
...
+        return self.seats_in_row * self.rows
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall, on_delete=models.CASCADE)
+    movie = models.ForeignKey(Movie, on_delete=models.CASCADE)
+
+    def __str__(self) -> str:
+        return f"{self.movie} {self.show_time.strftime('%Y-%m-%d %H:%M:%S')}"
```
YuriiKindrat, line 48: no need to use strftime here

services/movie_session.py
```diff
@@ -0,0 +1,56 @@
...
+from django.db.models import QuerySet
+
+from db.models import MovieSession, CinemaHall, Movie
+
+
+def create_movie_session(
+    movie_show_time: datetime, movie_id: int, cinema_hall_id: int
+) -> None:
+    cinema_hall = CinemaHall.objects.get(id=cinema_hall_id)
+    movie = Movie.objects.get(id=movie_id)
```
YuriiKindrat, line 12: use lookups instead

services/movie_session.py
YuriiKindrat: no need to use all here, u can use filter as below

services/movie_session.py
YuriiKindrat: same

YuriiKindrat (changes requested):

wQuelS (approved):
Good

PR #264, opened 2023-03-10
Solution

ZhAlexR:


services/cinema_hall.py
```diff
@@ -0,0 +1,24 @@
...
+
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> QuerySet:
+    return CinemaHall.objects.all()
+
+
+def get_cinema_hall_by_id(hall_id: int) -> QuerySet:
+    return CinemaHall.objects.get(hall_id)
```
SergiiMachulin, line 11: .get() method does not return a QuerySet

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
+from django.db.models import QuerySet
+from db.models import MovieSession
+
+
+def create_movie_session(
+        movie_show_time: str,
+        cinema_hall_id: int,
+        movie_id: int
+
+) -> MovieSession:
```
SergiiMachulin, line 10: You don't have any return inside, but the annotation is present.

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
...
+    return MovieSession.objects.get(id=movie_sessions_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: str = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None,
+
+) -> MovieSession:
```
SergiiMachulin, line 37: The same with annotation and return here.

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
...
+        movie_session.movie_id = movie_id
+
+    if cinema_hall_id:
+        movie_session.cinema_hall_id = cinema_hall_id
+
+    movie_session.save()
+
+
+def delete_movie_session_by_id(session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=session_id).delete()
```
SergiiMachulin, line 53: It is not a good idea and a practice to return something from a delete() method, because it simply deletes the object(s) from the database.

SergiiMachulin (changes requested):

services/movie_session.py
ZhAlexR: Why if I want to show the info about deleted object? Something like pop() method works in pure python ...

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
+from django.db.models import QuerySet
+from db.models import MovieSession
+
+
+def create_movie_session(
+        movie_show_time: str,
+        cinema_hall_id: int,
+        movie_id: int
+
```
YuriiKindrat, line 9: unnecessary blank line here

services/movie_session.py
YuriiKindrat: are u sure that delete method returns smth ?

services/movie.py
```diff
@@ -0,0 +1,41 @@
...
+        title=movie_title,
+        description=movie_description
+    )
+    if genres_ids:
+        movie.genres.set(genres_ids)
+
+    if actors_ids:
+        movie.actors.set(actors_ids)
+
+    return movie
```
wQuelS, line 41: you don't need to return something here

services/movie_session.py
```diff
@@ -0,0 +1,51 @@
+from django.db.models import QuerySet
+from db.models import MovieSession
+
+
+def create_movie_session(
+        movie_show_time: str,
+        cinema_hall_id: int,
+        movie_id: int
+) -> MovieSession:
+    return MovieSession.objects.create(
```
wQuelS, line 10: same

services/movie_session.py
```diff
@@ -0,0 +1,51 @@
...
+    if show_time:
+        movie_session.show_time = show_time
+
+    if movie_id:
+        movie_session.movie_id = movie_id
+
+    if cinema_hall_id:
+        movie_session.cinema_hall_id = cinema_hall_id
+
+    return movie_session.save()
```
wQuelS, line 47: same

wQuelS (changes requested):

YuriiKindrat (approved):

PR #270, opened 2023-03-12
'Solution'

alexlega:


services/movie.py
```diff
@@ -0,0 +1,41 @@
...
+        movies = movies.filter(actors__id__in=actors_ids)
+
+    return movies
+
+
+def get_movie_by_id(movie_id: int) -> QuerySet:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
```
ykopatko, line 24: When new movie is created, the result might be Movie instance.

services/movie_session.py
```diff
@@ -0,0 +1,47 @@
+from db.models import MovieSession
+from datetime import datetime
+from django.db.models import QuerySet
+
+
+def create_movie_session(
+        movie_show_time: datetime,
+        movie_id: int,
+        cinema_hall_id: int
+) -> None:
```
ykopatko, line 10: the same

services/cinema_hall.py
```diff
@@ -0,0 +1,18 @@
+from db.models import CinemaHall
+from django.db.models import QuerySet
+
+
+def get_cinema_halls() -> QuerySet:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(
```
ykopatko, line 9: When we create CinemaHall, we might return CinemaHall instance.

ykopatko (approved):

services/cinema_hall.py
YuriiKindrat: @ykopatko is right fix ur type annotation

services/movie.py
YuriiKindrat: same

services/movie_session.py
YuriiKindrat: same

YuriiKindrat (changes requested):

PR #272, opened 2023-03-12
solution

Yulia-Mala:


services/cinema_hall.py
```diff
@@ -0,0 +1,16 @@
...
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(hall_name: str,
+                       hall_rows: int,
+                       hall_seats_in_row: int) -> CinemaHall:
+    new_hall = CinemaHall.objects.create(name=hall_name,
+                                         rows=hall_rows,
+                                         seats_in_row=hall_seats_in_row)
+    return new_hall
```
YuriiKindrat, line 16: no need to create variable here

services/movie.py
```diff
@@ -0,0 +1,34 @@
...
+                                     description=movie_description)
+    try:
+        if genres_ids:
+            new_movie.genres.set(genres_ids)
+        if actors_ids:
+            new_movie.actors.set(actors_ids)
+    except Exception:
+        new_movie.delete()
+        return False
+    return new_movie
```
YuriiKindrat, line 34: it is bad practice to use Exception for exceptions catching 
also, fix this method according to requirements 
image
genres/actors in optional args, method should create movie in any case

services/movie_session.py
```diff
@@ -0,0 +1,42 @@
...
+from datetime import datetime
+
+
+def create_movie_session(movie_show_time: datetime,
+                         movie_id: int,
+                         cinema_hall_id: int) -> MovieSession:
+    new_session = MovieSession.objects.create(show_time=movie_show_time,
+                                              cinema_hall_id=cinema_hall_id,
+                                              movie_id=movie_id)
+    return new_session
```
YuriiKindrat, line 13: unnecessary variable

YuriiKindrat (changes requested):

services/movie_session.py
```diff
@@ -0,0 +1,41 @@
+from django.db.models import QuerySet
+
+from db.models import MovieSession
+from datetime import datetime
```
wQuelS, line 4: fix imports according to pep8

services/movie_session.py
```diff
@@ -0,0 +1,41 @@
...
+def create_movie_session(movie_show_time: datetime,
+                         movie_id: int,
+                         cinema_hall_id: int) -> MovieSession:
+    return MovieSession.objects.create(show_time=movie_show_time,
+                                       cinema_hall_id=cinema_hall_id,
+                                       movie_id=movie_id)
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
+    result = MovieSession.objects.all()
```
wQuelS, line 16: `queryset` as a name for this variable would be more descriptive

wQuelS (changes requested):

wQuelS (approved):
Great

PR #274, opened 2023-03-12
Solution

oigod:


db/models.py
```diff
@@ -14,3 +14,36 @@ class Actor(models.Model):
...
+        return self.rows * self.seats_in_row
+
+    def __str__(self) -> str:
+        return self.name
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall, on_delete=models.PROTECT)
+    movie = models.ForeignKey(Movie, on_delete=models.PROTECT)
```
YuriiKindrat, line 45: why do you use PROTECT here?

db/models.py
```diff
@@ -14,3 +14,36 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
```

db/models.py
```diff
@@ -14,3 +14,36 @@ class Actor(models.Model):
...
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall, on_delete=models.PROTECT)
+    movie = models.ForeignKey(Movie, on_delete=models.PROTECT)
+
+    def __str__(self) -> str:
+        show_time_str = self.show_time.strftime("%Y-%m-%d %H:%M:%S")
+        return str(self.movie) + " " + show_time_str
```
YuriiKindrat, line 49: no need to use strftime here

services/cinema_hall.py
```diff
@@ -0,0 +1,20 @@
+from django.db.models import QuerySet
+
+
```
YuriiKindrat, line 3: unnecessary blank line

services/movie.py
```diff
@@ -0,0 +1,45 @@
...
+
+    if genres_ids and actors_ids:
+        return Movie.objects.filter(genres__id__in=genres_ids,
+                                    actors__id__in=actors_ids
+                                    )
+    if genres_ids and actors_ids is None:
+        return Movie.objects.filter(genres__id__in=genres_ids)
+
+    if actors_ids and genres_ids is None:
+        return Movie.objects.filter(actors__id__in=actors_ids)
```
YuriiKindrat, line 21: try to simplifize this, u can use only 2 conditions here, also fix indentation

services/movie_session.py
```diff
@@ -0,0 +1,47 @@
...
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        # show_time: datetime = None,
+        # movie_id: int = None,
+        # cinema_hall_id: int = None
```
YuriiKindrat, line 38: remove unnecessary commnets here

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
...
+        return self.name
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall, on_delete=models.SET_NULL)
+    movie = models.ForeignKey(Movie, on_delete=models.SET_NULL)
+
+    def __str__(self) -> str:
+        return str(self.movie) + " " + str(self.show_time)
```
wQuelS, line 48: use f-string. It is faster and more reliable than concatenation

services/cinema_hall.py
```diff
@@ -0,0 +1,19 @@
...
+def get_cinema_halls() -> QuerySet:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(
+        hall_name: str,
+        hall_rows: int,
+        hall_seats_in_row: int
+) -> CinemaHall:
+    return CinemaHall.objects.create(
```
wQuelS, line 15: The task was to create `Cinemahall`, you don't actually need to return it

services/movie_session.py
```diff
@@ -0,0 +1,44 @@
...
+def create_movie_session(
+        movie_show_time: datetime,
+        movie_id: int,
+        cinema_hall_id: int
+) -> MovieSession:
+    return MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
```
wQuelS, line 17: not needed to `return` anything here

services/movie_session.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def get_movies_sessions(
+        session_date: str = None
+) -> QuerySet:
+    if session_date:
+        return MovieSession.objects.filter(
+            show_time__date=session_date
+        )
+    return MovieSession.objects.all()
```
wQuelS, line 27: You can store all `MovieSession` in a variable and if the condition in line 23 is `True` -> apply needed filters. In such approach you can avoid using two `returns`

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
...
+        return self.rows * self.seats_in_row
+
+    def __str__(self) -> str:
+        return self.name
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall, on_delete=models.SET_NULL)
+    movie = models.ForeignKey(Movie, on_delete=models.SET_NULL)
```
wQuelS, line 45: `SET_NULL` is not correct either

services/movie_session.py
```diff
@@ -0,0 +1,44 @@
...
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        **kwargs
+) -> None:
+    return MovieSession.objects.filter(
+        id=session_id
+    ).update(**kwargs)
```
wQuelS, line 40: You don't need to `return` here

wQuelS (changes requested):

services/movie_session.py
oigod: i think it was a good practice because of how method create work so i wrote returns 
image

oigod: am i wrong?

wQuelS: Of course ORM's method `create` returns an object that was created, because you sometimes need something like that: `var = Model.objects.create(...)`. But you are not implementing `create` method for framework's ORM :)
All you need is to have method that will create MoviSession record in db, that' it

wQuelS (approved):
Great

services/movie_session.py
oigod: got u bro <3

oigod: thx a lot for explain

PR #275, opened 2023-03-12
Solution

tytomyr:


skmiwa1 (approved):

db/models.py
```diff
@@ -14,3 +14,40 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
```
wQuelS, line 23: you should add `related_name`, since it is according to best practices

db/models.py
```diff
@@ -14,3 +14,40 @@ class Actor(models.Model):
...
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
+
+    def __str__(self) -> str:
+        return f"{self.name}"
+
+    @property
+    def capacity(self) -> int:
+        capacity = self.rows * self.seats_in_row
+        return capacity
```
wQuelS, line 40: you could just return that simple calculation

db/models.py
```diff
@@ -14,3 +14,40 @@ class Actor(models.Model):
...
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall,
+                                    on_delete=models.SET_NULL,
+                                    null=True)
+    movie = models.ForeignKey(Movie,
+                              on_delete=models.SET_NULL,
+                              null=True)
```
wQuelS, line 50: Why do you use `SET_NULL` here ?

services/cinema_hall.py
```diff
@@ -0,0 +1,16 @@
...
+
+def create_cinema_hall(hall_name: str,
+                       hall_rows: int,
+                       hall_seats_in_row: int) -> CinemaHall:
+    hall = CinemaHall.objects.create(
+        name=hall_name,
+        rows=hall_rows,
+        seats_in_row=hall_seats_in_row
+    )
+    return hall
```
wQuelS, line 16: you don't need additional variable here. Also `return` is not needed too

wQuelS (changes requested):

db/models.py
```diff
@@ -14,3 +14,39 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="performers")
+    genres = models.ManyToManyField(Genre, related_name="category")
```
YuriiKindrat, line 23: incorrect related names - check ur checklist
are u sure about `actor.performers.all()` should return all movies with actor ?

db/models.py
```diff
@@ -14,3 +14,39 @@ class Actor(models.Model):
...
+
+    def __str__(self, description: str = None) -> str:
+        return f"{self.title}"
```
YuriiKindrat, line 26: no need to use f string here

db/models.py
```diff
@@ -14,3 +14,39 @@ class Actor(models.Model):
...
+
+
+class CinemaHall(models.Model):
+    name = models.CharField(max_length=255)
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
+
+    def __str__(self) -> str:
+        return f"{self.name}"
```
YuriiKindrat, line 35: same

db/models.py
```diff
@@ -14,3 +14,39 @@ class Actor(models.Model):
...
+        return self.rows * self.seats_in_row
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall,
+                                    on_delete=models.DO_NOTHING,
+                                    null=True)
+    movie = models.ForeignKey(Movie,
+                              on_delete=models.DO_NOTHING,
```
YuriiKindrat, line 48: why do u use do nothing here, co if we'll delete a movie/cinemahall, movie session still can exist but without movie and cinema hall?

services/movie.py
```diff
@@ -0,0 +1,35 @@
+from db.models import Movie
+
+
+def get_movies(genres_ids: list = None,
+               actors_ids: list = None) -> Movie:
```
YuriiKindrat, line 5: are u sure that this method returns Movie instance ?

services/movie_session.py
```diff
@@ -0,0 +1,47 @@
...
+                         movie_id: int,
+                         cinema_hall_id: int) -> MovieSession:
+    return MovieSession.objects.create(
+        show_time=movie_show_time,
+        cinema_hall_id=cinema_hall_id,
+        movie_id=movie_id
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> MovieSession:
```
YuriiKindrat, line 15: same

services/movie_session.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(session_id: int,
+                         show_time: str = None,
+                         movie_id: int = None,
+                         cinema_hall_id: int = None) -> MovieSession:
```
YuriiKindrat, line 29: same

YuriiKindrat (changes requested):

services/movie.py
```diff
@@ -0,0 +1,37 @@
...
+        description=movie_description
+    )
+
+    if genres_ids:
+        movie.genres.set(genres_ids)
+
+    if actors_ids:
+        movie.actors.set(actors_ids)
+
+    return movie
```
wQuelS, line 37: you actually don't need return here

wQuelS (changes requested):

PR #276, opened 2023-03-12
Solution

oksana0410:


db/models.py
```diff
@@ -14,3 +15,37 @@ class Actor(models.Model):
...
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
```
YuriiKindrat, line 25: it is better to use related names for relations

services/cinema_hall.py
```diff
@@ -0,0 +1,19 @@
+from db.models import CinemaHall
+
+from django.db.models.query import QuerySet
```
YuriiKindrat, line 3: fix imports according to pep8

services/movie.py
```diff
@@ -0,0 +1,39 @@
+from db.models import Movie
+
+from django.db.models.query import QuerySet
```
YuriiKindrat, line 3: same

services/movie_session.py
```diff
@@ -0,0 +1,47 @@
+from db.models import MovieSession
+
+from django.db.models.query import QuerySet
```
YuriiKindrat, line 3: same

PR #278, opened 2023-03-12
Solution

ada-krav:


services/movie_session.py
```diff
@@ -0,0 +1,51 @@
...
+    movie_id: int,
+    cinema_hall_id: int
+) -> MovieSession:
+
+    new_movie_session = MovieSession.objects.create(
+        show_time=movie_show_time,
+        cinema_hall_id=cinema_hall_id,
+        movie_id=movie_id
+    )
+    return new_movie_session
```
YuriiKindrat, line 19: no need to create variable here

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
```

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #281, opened 2023-03-12
Solution

Danil228:


services/movie_session.py
```diff
@@ -0,0 +1,45 @@
...
+                         movie_id: int,
+                         cinema_hall_id: int) -> None:
+    MovieSession.objects.create(
+        show_time=movie_show_time,
+        cinema_hall_id=cinema_hall_id,
+        movie_id=movie_id
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet | MovieSession:
```
YuriiKindrat, line 18: are u sure that filter method returns moviesession ?

YuriiKindrat (changes requested):

wQuelS (changes requested):

wQuelS (approved):
Good

PR #282, opened 2023-03-13
Solutionis

skmiwa1:


db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
...
+        return self.rows * self.seats_in_row
+
+    def __str__(self) -> str:
+        return self.name
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall, on_delete=models.CASCADE, related_name="movies")
+    movie = models.ForeignKey(Movie, on_delete=models.CASCADE, related_name="movies")
```
YuriiKindrat, line 45: are u sure about movies related name? maybe it is better to use movie_sessions or sessions?

services/movie_session.py
```diff
@@ -0,0 +1,51 @@
...
+    movie_id: int,
+    cinema_hall_id: int
+) -> MovieSession:
+
+    new_movie_session = MovieSession.objects.create(
+        show_time=movie_show_time,
+        cinema_hall_id=cinema_hall_id,
+        movie_id=movie_id
+    )
+    return new_movie_session
```
YuriiKindrat, line 19: no need to create variable here

YuriiKindrat (changes requested):
also fix ur flake8 errors

PR #283, opened 2023-03-13
Solution

Zenionoi:


db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
```

YuriiKindrat (changes requested):

db/models.py
```diff
@@ -14,3 +14,39 @@ class Actor(models.Model):
...
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(
+        CinemaHall, on_delete=models.CASCADE, related_name="session"
+    )
+    movie = models.ForeignKey(
+        Movie, on_delete=models.CASCADE, related_name="session"
+    )
```
wQuelS, line 49: `related_name` should be in plural, try to guess why)

services/movie_session.py
```diff
@@ -0,0 +1,51 @@
...
+    MovieSession.objects.create(
+        show_time=movie_show_time,
+        cinema_hall_id=cinema_hall_id,
+        movie_id=movie_id
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
+    return (MovieSession.objects.filter(show_time__date=session_date)
+            if session_date else MovieSession.objects.all())
```
wQuelS, line 22: not all oneliners are good, make it more readable, but with one return

wQuelS (changes requested):

YuriiKindrat (approved):

PR #284, opened 2023-03-13
'Solution'

NickSavchyn:


db/models.py
```diff
@@ -14,3 +14,39 @@ class Actor(models.Model):
...
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(
+        CinemaHall, on_delete=models.CASCADE,
+    )
+    movie = models.ForeignKey(
+        Movie, on_delete=models.CASCADE,
+    )
+
+    def __str__(self) -> str:
+        return f"{self.movie} {str(self.show_time)}"
```
YuriiKindrat, line 52: no need to use str here)

services/movie.py
```diff
@@ -0,0 +1,43 @@
...
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: Optional[list[int]] = None,
+        actors_ids: Optional[list[int]] = None
+) -> None:
```
YuriiKindrat, line 32: are u sure that u returns None here ?

services/movie_session.py
```diff
@@ -0,0 +1,55 @@
...
+
+from django.db.models import QuerySet
+
+from db.models import MovieSession
+
+
+def create_movie_session(
+        movie_show_time: str,
+        movie_id: int,
+        cinema_hall_id: int) -> QuerySet:
```
YuriiKindrat, line 11: fix indentation here

YuriiKindrat (changes requested):

PR #287, opened 2023-03-13
'Solution'

Bohdan-Kovalenko1992:


AlessaQ (approved):

services/movie_session.py
```diff
@@ -0,0 +1,55 @@
...
+
+def create_movie_session(
+        movie_show_time: datetime,
+        movie_id: int,
+        cinema_hall_id: int
+) -> None:
+    MovieSession.objects.create(
+        show_time=movie_show_time,
+        cinema_hall=CinemaHall.objects.get(id=cinema_hall_id),
+        movie=get_movie_by_id(movie_id)
```
YuriiKindrat, line 17: no need to use get methods here, use lookups instead

services/movie_session.py
```diff
@@ -0,0 +1,55 @@
...
+    )
+
+
+def get_movies_sessions(session_date: (list[str], None) = None) -> QuerySet:
+    queryset = MovieSession.objects.all()
+    if session_date:
+        queryset = queryset.filter(
+            show_time__date=datetime.strptime(
```
YuriiKindrat, line 25: no need to use strptime here

YuriiKindrat (changes requested):

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
```
YuriiKindrat, line 23: also it is better to use related names

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
...
+        return self.rows * self.seats_in_row
+
+    def __str__(self) -> str:
+        return self.name
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall, on_delete=models.CASCADE)
+    movie = models.ForeignKey(Movie, on_delete=models.CASCADE)
```
wQuelS, line 45: `related_names` is almost always a good practice. Please, don't forget about it

wQuelS (changes requested):

Nattalli (approved):

PR #288, opened 2023-03-13
Solution

Yevheniy-Lototskiy:


services/movie_session.py
```diff
@@ -0,0 +1,63 @@
...
+
+    if session_date:
+        year, month, day = session_date.split("-")
+        movie_session = movie_session.filter(
+            show_time__date=datetime.date(
+                year=int(year),
+                month=int(month),
+                day=int(day)
+            )
+        )
```
YuriiKindrat, line 31: no need to use split and datetime.date here

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
```
YuriiKindrat, line 23: it is better to use related_name arg for relations

YuriiKindrat (changes requested):

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="actors")
+    genres = models.ManyToManyField(Genre, related_name="genres")
```
YuriiKindrat, line 23: incorrect names for related_name args check ur checklist

YuriiKindrat (changes requested):

wQuelS (approved):
Good

PR #292, opened 2023-03-14
'Solution'

Vitalii2906:


services/movie.py
```diff
@@ -0,0 +1,52 @@
+from typing import List
+
+from db.models import Genre, Movie, Actor
+
+
+def get_movies(
+        genres_ids: List[Genre] = None,
```
vasylhnatiuk, line 7: add Optional here and other place where argument can be None

services/movie.py
```diff
@@ -0,0 +1,52 @@
...
+        return queryset
+
+    if actors_ids:
+        queryset = queryset.filter(
+            actors__id__in=actors_ids
+        )
+        return queryset
+
+
+def get_movie_by_id(movie_id: int) -> List[Movie]:
```
vasylhnatiuk, line 35: this func returns only one Movie, not a list

vasylhnatiuk (changes requested):

Grayder0152 (changes requested):
Pls resolve comments

Vitalii2906:
i fixed it

services/movie_session.py
```diff
@@ -0,0 +1,64 @@
...
+def update_movie_session(
+        session_id: int,
+        show_time: str = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> None:
+    try:
+        movie = Movie.objects.get(id=movie_id)
+    except Movie.DoesNotExist:
+        movie = None
```
ArsenPidhoretskyi, line 40: Use get_object_or_404

Vitalii2906:
I don't understand what I need to do or what exactly I didn't fix

ArsenPidhoretskyi (changes requested):

services/movie.py
Vitalii2906: "Second, it also Optional because default parameter is None"
I can't understand what it means

services/movie.py
```diff
@@ -0,0 +1,41 @@
+from typing import List
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: List[int] = None,
+        actors_ids: List[int] = None
```
katevasilenko, line 8: https://stackoverflow.com/questions/51710037/how-should-i-use-the-optional-type-hint
Please get familiar with Optional and fix all over the project

services/movie_session.py
```diff
@@ -0,0 +1,59 @@
...
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: str = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> None:
+    queryset = MovieSession.objects.filter(id=session_id)
```
katevasilenko, line 40: You can also use `get_movie_session_by_id ` here

katevasilenko (changes requested):

services/movie.py
Vitalii2906: thank you very much)

services/movie.py
```diff
@@ -0,0 +1,41 @@
+from typing import Optional, List
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: Optional[int] = None,
+        actors_ids: Optional[int] = None
```
ArsenPidhoretskyi, line 8: No it is list of int not one integer
Optional[list[int]]

ArsenPidhoretskyi: Fix in other places

services/movie_session.py
```diff
@@ -0,0 +1,60 @@
...
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> None:
+    movie_session = get_movie_session_by_id(session_id)
+
+    try:
+        movie = get_object_or_404(Movie, pk=movie_id)
+        movie_session.movie = movie
+    except Http404:
+        pass
```
ArsenPidhoretskyi, line 46: Why you need to get movie object? Learn how to set foreign key value by id
The same with cinema_hall

ArsenPidhoretskyi (changes requested):

services/movie_session.py
```diff
@@ -0,0 +1,49 @@
...
+    return MovieSession.objects.filter(show_time__date=session_date)
+
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: str = None,
```
vasylhnatiuk, line 33: Use Optional here as well

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):
Good job!

PR #293, opened 2023-03-15
Solution

StreetNik:


services/movie.py
```diff
@@ -0,0 +1,41 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+    genres_ids: list[int] = None,
+    actors_ids: list[int] = None
```
ArsenPidhoretskyi, line 8: It accepts Optional[list[int]] because default parameter already None
Fix in other places

services/movie_session.py
```diff
@@ -0,0 +1,62 @@
...
+        movie_show_time: str = None
+) -> QuerySet:
+    query_set = MovieSession.objects.filter()
+    if movie_show_time:
+        movie_show_time = movie_show_time.split("-")
+        query_set = query_set.filter(
+            show_time__year=movie_show_time[0],
+            show_time__month=movie_show_time[1],
+            show_time__day=movie_show_time[2]
+        )
```
ArsenPidhoretskyi, line 31: Simplify this

> Note: You can use suffix __date to get access to date of the DateTimeField

ArsenPidhoretskyi (changes requested):

services/movie_session.py
```diff
@@ -0,0 +1,61 @@
...
+    return query_set
+
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: Optional[datetime] = None,
```
ArsenPidhoretskyi: datetime is module
Use datetime.datetime here

ArsenPidhoretskyi (changes requested):

services/movie_session.py
StreetNik: I did it in last commit

services/movie_session.py
```diff
@@ -0,0 +1,61 @@
+from typing import Optional
+
+from db.models import MovieSession
+from django.db.models import QuerySet
+
+import datetime
+
+
+def create_movie_session(
+        movie_show_time: datetime,
```
ArsenPidhoretskyi, line 10: Here is also datetime.datetime
Please do self-review.....
You do:
import datetime
that means datetime is module

ArsenPidhoretskyi (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #295, opened 2023-03-16
Solution

nazar-dmytrenko:


services/movie.py
```diff
@@ -0,0 +1,50 @@
...
+        genres_ids: Optional[list[int]] = None,
+        actors_ids: Optional[list[int]] = None
+) -> QuerySet[Movie]:
+    queryset = Movie.objects.all()
+
+    if genres_ids is not None and actors_ids is not None:
+        queryset = queryset.filter(
+            genres__id__in=genres_ids,
+            actors__id__in=actors_ids
+        )
```
ArsenPidhoretskyi, line 18: It is not needed because you already covers it in the next lines

services/movie_session.py
```diff
@@ -0,0 +1,58 @@
...
+) -> None:
+
+    MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
+
+
+def get_movies_sessions(session_date: str = None,) -> QuerySet[MovieSession]:
```
ArsenPidhoretskyi, line 22: You forgot about Optional[str]

ArsenPidhoretskyi (changes requested):

PR #301, opened 2023-03-20
Solution

6aPa6aH:


services/movie.py
```diff
@@ -0,0 +1,36 @@
+from db.models import Movie
+from django.db.models import QuerySet
+
+
+def get_movies(genres_ids: list[int] = None,
+               actors_ids: list[int] = None) -> QuerySet:
```
ArsenPidhoretskyi, line 6: It accepts Optional[list[int]] because have as default None
Fix in other places

services/movie.py
```diff
@@ -0,0 +1,36 @@
...
+    if genres_ids is not None:
+        queryset = queryset.filter(genres__id__in=genres_ids)
+
+    if actors_ids is not None:
+        queryset = queryset.filter(actors__id__in=actors_ids)
+
+    return queryset
+
+
+def get_movie_by_id(movie_id: int) -> QuerySet:
```
ArsenPidhoretskyi, line 18: It return Movie, not queryset

services/movie_session.py
```diff
@@ -0,0 +1,47 @@
...
+
+def get_movies_sessions(session_date: datetime = None) -> QuerySet:
+    queryset = MovieSession.objects.all()
+
+    if session_date:
+        queryset = queryset.filter(show_time__date=session_date)
+    return queryset
+
+
+def get_movie_session_by_id(movie_session_id: int) -> QuerySet:
```
ArsenPidhoretskyi, line 25: It returns MovieSession

services/movie_session.py
```diff
@@ -0,0 +1,47 @@
...
+    queryset = MovieSession.objects.get(id=movie_session_id)
+    return queryset
+
+
+def update_movie_session(session_id: int,
+                         show_time: datetime = None,
```
ArsenPidhoretskyi, line 31: It accepts module?

ArsenPidhoretskyi (changes requested):

.github/__init__.py
ArsenPidhoretskyi, line 1: Why you need this file???

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
...
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
+
+    def __str__(self) -> str:
+        return f"{self.title}"
```
ArsenPidhoretskyi, line 26: f-string is useless here

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
...
+
+
+class CinemaHall(models.Model):
+    name = models.CharField(max_length=255)
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
+
+    def __str__(self) -> str:
+        return f"{self.name}"
```
ArsenPidhoretskyi, line 35: f-string is useless here

services/cinema_hall.py
```diff
@@ -0,0 +1,15 @@
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> CinemaHall:
```
ArsenPidhoretskyi, line 4: First of all it not return one object
Also it does not return objects, it return QuerySet

services/movie.py
```diff
@@ -0,0 +1,38 @@
...
+        queryset = queryset.filter(genres__id__in=genres_ids)
+
+    if actors_ids is not None:
+        queryset = queryset.filter(actors__id__in=actors_ids)
+
+    return queryset
+
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    queryset = Movie.objects.get(id=movie_id)
```
ArsenPidhoretskyi, line 21: queryset variable name is wrong
rename it to movie, because .get return object

services/movie_session.py
```diff
@@ -0,0 +1,48 @@
...
+    queryset = MovieSession.objects.all()
+
+    if session_date:
+        queryset = queryset.filter(show_time__date=session_date)
+    return queryset
+
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    queryset = MovieSession.objects.get(id=movie_session_id)
+    return queryset
```
ArsenPidhoretskyi, line 28: The same as with Movie

services/movie_session.py
```diff
@@ -0,0 +1,48 @@
...
+                         cinema_hall_id: Optional[int] = None) -> None:
+    queryset = MovieSession.objects.filter(id=session_id)
+    if show_time:
+        queryset.update(show_time=show_time)
+
+    if movie_id:
+        queryset.update(movie_id=movie_id)
+
+    if cinema_hall_id:
+        queryset.update(cinema_hall_id=cinema_hall_id)
```
ArsenPidhoretskyi, line 43: YOu can do it in 2 queries. No need to use .update because each .update do 1 query

ArsenPidhoretskyi (changes requested):

services/movie.py
```diff
@@ -0,0 +1,38 @@
...
+def create_movie(movie_title: str,
+                 movie_description: str,
+                 genres_ids: Optional[list[int]] = None,
+                 actors_ids: Optional[list[int]] = None) -> None:
+    new_film = Movie.objects.create(
+        title=movie_title,
+        description=movie_description
+    )
+
+    if genres_ids:
```
Grayder0152, line 34:
```suggestion
    if genres_ids is not None:
```

Grayder0152 (approved):

PR #302, opened 2023-03-21
Solution

destr34:


services/movie.py
```diff
@@ -0,0 +1,36 @@
+from db.models import Movie
+
+from django.db.models import QuerySet
+
+
+def get_movies(
+        genres_ids: str = None,
```
ArsenPidhoretskyi, line 7: First of all, it accepts list of ints. The same with actors_ids
Also it will be optional because accepts None(fix this case for all code)

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
+from db.models import MovieSession
+
+from django.db.models import QuerySet
+
+from typing import Optional
+import datetime
+
+
+def create_movie_session(
+        movie_show_time: Optional,
```
ArsenPidhoretskyi, line 10: Optional what?
Provide inner type hint

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+    return queryset
+
+
+def get_movie_session_by_id(session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: datetime = None,
```
ArsenPidhoretskyi, line 34: It accepts module? 
Because for now you import:
import datetime that mean datetime is module
Do:
`from datetime import datetime`
Or change typing to datetime.datetime

ArsenPidhoretskyi (changes requested):

services/movie.py
```diff
@@ -0,0 +1,38 @@
+from db.models import Movie
+
+from django.db.models import QuerySet
+from typing import List
+
+
+def get_movies(
+        genres_ids: List[int] = None,
+        actors_ids: List[int] = None
```
Bohdan-Salamakha, line 9: use typehint like: Optional[YOUR_VALUE_TYPE] = None

services/movie.py
```diff
@@ -0,0 +1,38 @@
...
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: List[int] = None,
+        actors_ids: List[int] = None
```
Bohdan-Salamakha, line 28: use typehint like: Optional[YOUR_VALUE_TYPE] = None

services/movie_session.py
```diff
@@ -0,0 +1,49 @@
...
```
Bohdan-Salamakha, line 33: use typehint like: Optional[YOUR_VALUE_TYPE] = None

services/movie_session.py
```diff
@@ -0,0 +1,49 @@
...
+        movie_id: int = None,
+        cinema_hall_id: int = None
```
Bohdan-Salamakha, line 35: use typehint like: Optional[YOUR_VALUE_TYPE] = None

services/movie.py
```diff
@@ -0,0 +1,38 @@
...
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: Optional[int] = None,
+        actors_ids: Optional[int] = None
```
ArsenPidhoretskyi, line 28: It accepts Optional[list[int]] not one int

ArsenPidhoretskyi (approved):
Well done!

PR #303, opened 2023-03-23
Solution

Rename925:


services/movie.py
```diff
@@ -0,0 +1,41 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(genres_ids: list[int] = None,
+               actors_ids: list[int] = None
```
ArsenPidhoretskyi, line 7: It accepts Optional[list[int]] because default parameter is None
Fix in other places

services/movie_session.py
```diff
@@ -0,0 +1,43 @@
...
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id,
+
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
+    if session_date:
+        return MovieSession.objects.all().filter(show_time__date=session_date)
+    return MovieSession.objects.all()
```
ArsenPidhoretskyi, line 21:
```suggestion
    queryset = MovieSession.objects.all()
    if session_date:
        queryset = queryset.filter(show_time__date=session_date)
    return queryset

```

ArsenPidhoretskyi (changes requested):

db/models.py
```diff
@@ -14,3 +14,36 @@ class Actor(models.Model):
...
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall, on_delete=models.CASCADE)
+    movie = models.ForeignKey(Movie, on_delete=models.CASCADE)
+
+    def __str__(self) -> str:
+        return f"{self.movie.title} " \
+               f"{self.show_time.strftime('%Y-%m-%d %H:%M:%S')}"
```
katevasilenko, line 49: You can use brackets to avoid \ symbol

services/movie_session.py
```diff
@@ -0,0 +1,46 @@
...
+        queryset = queryset.filter(show_time__date=session_date)
+    return queryset
+
+
+def get_movie_session_by_id(movie_session_id: int) -> QuerySet:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(session_id: int,
+                         show_time: str = None,
```
katevasilenko, line 32: I believe you need to use Optional here too

services/cinema_hall.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+def get_cinema_halls() -> QuerySet:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(hall_name: str,
+                       hall_rows: int,
+                       hall_seats_in_row: int
+                       ) -> QuerySet:
```
katevasilenko, line 13:
```suggestion

def create_cinema_hall(
         hall_name: str,
         hall_rows: int,
         hall_seats_in_row: int
 ) -> None:
```
This style with splitting line looks neater for me. And you do not return anything in this function

katevasilenko (changes requested):

vasylhnatiuk (approved):
Well done!

PR #308, opened 2023-03-30
Add solution

ZaikaDaria:


services/movie.py
```diff
@@ -0,0 +1,46 @@
...
+    if genres_ids and actors_ids:
+        movies = Movie.objects.filter(
+            genres__id__in=genres_ids,
+            actors__id__in=actors_ids
+        )
+
+    if genres_ids and not actors_ids:
+        movies = Movie.objects.filter(genres__id__in=genres_ids)
+    if actors_ids and not genres_ids:
+        movies = Movie.objects.filter(actors__id__in=actors_ids)
```
ArsenPidhoretskyi, line 21: You can simplify it into two conditions:
1. if genres_ids is not None
2. if actors_ids is not None
you can do:
movies = movies.filter....

services/movie.py
```diff
@@ -0,0 +1,46 @@
...
+
+    return movies
+
+
+def get_movie_by_id(movie_id: int = None) -> Movie:
+    return Movie.objects.get(id=movie_id)
```
ArsenPidhoretskyi, line 27: Use get_object_or_404
It more practicable in this case

services/movie_session.py
```diff
@@ -0,0 +1,49 @@
...
+        cinema_hall_id=cinema_hall_id,
+        movie_id=movie_id
+    )
+    return new_movie_session
+
+
+def get_movies_sessions(session_date: Optional[str] = None) -> QuerySet:
+    scheduled_movies = MovieSession.objects.all()
+    if session_date is not None:
+        scheduled_movies = MovieSession.objects.filter(
```
ArsenPidhoretskyi, line 22: You can reuse your query:
```suggestion
        scheduled_movies = scheduled_movies.filter(
```

services/movie_session.py
```diff
@@ -0,0 +1,49 @@
...
+            show_time__date=session_date
+        )
+    return scheduled_movies
+
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    movies_by_session = MovieSession.objects.get(id=movie_session_id)
```
ArsenPidhoretskyi, line 29: Use get_object_or_404
If object was not found it will raise 404 exception

ArsenPidhoretskyi (changes requested):

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
```
vasylhnatiuk, line 22: add related_name

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
...
+    genres = models.ManyToManyField(Genre)
```
vasylhnatiuk, line 23: the same here

vasylhnatiuk (changes requested):

PR #309, opened 2023-03-31
Develop

U-Ra-All:


services/movie_session.py
```diff
@@ -0,0 +1,65 @@
...
+    )
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: Optional[str] = None,
+        movie_id: Optional[int] = None,
+        cinema_hall_id: Optional[int] = None
+) -> int:
+    queryset = MovieSession.objects.filter(id=session_id)
```
ArsenPidhoretskyi, line 47: just get the object and then update it by for example:
movie_sesion.some_property = new_value

services/movie_session.py
```diff
@@ -0,0 +1,65 @@
...
+
+        if cinema_hall_id:
+            updated_entries_number\
+                += queryset.update(cinema_hall_id=cinema_hall_id)
+
+    return updated_entries_number
+
+
+def delete_movie_session_by_id(session_id: int) -> int:
+    return MovieSession.objects.filter(id=session_id).delete()
```
ArsenPidhoretskyi, line 65: reuse get_movie_session_by_id

ArsenPidhoretskyi (changes requested):

services/movie_session.py
U-Ra-All: Done

Bohdan-Salamakha (approved):

PR #311, opened 2023-04-03
Solution v_01

bohdan-yatsyna:


services/movie_session.py
```diff
@@ -0,0 +1,49 @@
...
+    if movie_id:
+        session_to_update.movie_id = movie_id
+    if cinema_hall_id:
+        session_to_update.cinema_hall_id = cinema_hall_id
+
+    session_to_update.save()
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    MovieSession.objects.get(id=session_id).delete()
```
Vasyl-Smutok, line 49: use `get_movie_session_by_id`

Vasyl-Smutok (changes requested):

services/movie_session.py
bohdan-yatsyna: Thank you for the remark.

services/movie.py
```diff
@@ -0,0 +1,38 @@
...
+
+def get_movie_by_id(movie_id: int) -> QuerySet:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
vasylhnatiuk, line 28:
```suggestion
    actors_ids: Optional[list[int]] = None

```

vasylhnatiuk: use Optional if value can be None, import Optional from typing. use here and in all of the code

vasylhnatiuk (changes requested):

services/movie.py
bohdan-yatsyna: Noted with thanks.

ArsenPidhoretskyi (approved):
Well done!

PR #315, opened 2023-04-07
Solution

svitlana-savikhina:


services/movie.py
```diff
@@ -0,0 +1,32 @@
+from db.models import Movie
+from django.db.models import QuerySet
+
+
+def get_movies(genres_ids: list[int] = None,
+               actors_ids: list[int] = None
```
Bohdan-Salamakha, line 6: use typehint like: Optional[YOUR_VALUE_TYPE] = None

services/movie.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(movie_title: str,
+                 movie_description: str,
+                 genres_ids: list[int] = None,
+                 actors_ids: list[int] = None
```
Bohdan-Salamakha, line 23: use typehint like: Optional[YOUR_VALUE_TYPE] = None

services/movie_session.py
```diff
@@ -0,0 +1,44 @@
...
+def create_movie_session(movie_show_time: datetime,
+                         movie_id: int,
+                         cinema_hall_id: int
+                         ) -> None:
+    MovieSession.objects.create(show_time=movie_show_time,
+                                movie_id=movie_id,
+                                cinema_hall_id=cinema_hall_id)
+
+
+def get_movies_sessions(session_date: datetime = None) -> QuerySet:
```
Bohdan-Salamakha, line 15: use typehint like: Optional[YOUR_VALUE_TYPE] = None

services/movie_session.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(session_id: int,
+                         show_time: datetime = None,
+                         movie_id: datetime = None,
+                         cinema_hall_id: int = None) -> None:
```
Bohdan-Salamakha, line 29: use typehint like: Optional[YOUR_VALUE_TYPE] = None

Bohdan-Salamakha (changes requested):
v1

Bohdan-Salamakha (approved):

PR #318, opened 2023-04-15
Solution

Roman28101:


services/movie.py
```diff
@@ -0,0 +1,34 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
ArsenPidhoretskyi, line 8: It accepts Optional[list[int]] because default arguments is None.
Fix for all types in all places

services/movie_session.py
```diff
@@ -0,0 +1,46 @@
...
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> None:
+    queryset = MovieSession.objects.filter(id=session_id)
+    if show_time is not None:
+        queryset.update(show_time=show_time)
+    if movie_id is not None:
+        queryset.update(movie_id=movie_id)
+    if cinema_hall_id is not None:
+        queryset.update(cinema_hall_id=cinema_hall_id)
```
ArsenPidhoretskyi, line 42: Better to do this in two queries:
1. get movie session
2. set all values and than save it

ArsenPidhoretskyi (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #319, opened 2023-04-17
Solution

olenaliuby:


services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+        movie_session.movie_id = movie_id
+
+    if cinema_hall_id:
+        movie_session.cinema_hall_id = cinema_hall_id
+
+    movie_session.save()
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    MovieSession.objects.get(id=session_id).delete()
```
vasylhnatiuk, line 54: use get_movie_session_by_id() here

vasylhnatiuk (changes requested):

services/movie.py
```diff
@@ -0,0 +1,45 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
Grayder0152, line 8: + Optional typing

services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: datetime = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> None:
```
Grayder0152, line 38: + Optional

services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+    movie_session = MovieSession.objects.get(id=session_id)
+
+    if show_time:
```
Grayder0152, line 41:
```suggestion
    if show_time is not None:
```

services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+        movie_session.show_time = show_time
+
+    if movie_id:
```
Grayder0152, line 44: the same as above

services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+        movie_session.movie_id = movie_id
+
+    if cinema_hall_id:
```
Grayder0152, line 47: the same as above

Grayder0152 (changes requested):

services/movie.py
olenaliuby: Do you mean make it like that `genres_ids: Optional[list[int]]` instead of  `genres_ids: list[int] = None`?

PR #320, opened 2023-04-18
v1

NVMBR13:


services/movie.py
```diff
@@ -0,0 +1,32 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(genres_ids: list[int] = None,
+               actors_ids: list[int] = None
```

services/movie.py
```diff
@@ -0,0 +1,32 @@
...
+    return movies_set
+
+
+def get_movie_by_id(movie_id: int) -> QuerySet:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(movie_title: str, movie_description: str,
+                 genres_ids: list[int] = None, actors_ids: list[int] = None
+                 ) -> None:
```
Bohdan-Salamakha, line 24: use typehint like: Optional[YOUR_VALUE_TYPE] = None

format it nicer

services/movie_session.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def create_movie_session(movie_show_time: datetime.datetime, movie_id: int,
+                         cinema_hall_id: int) -> None:
+    MovieSession.objects.create(show_time=movie_show_time,
+                                movie_id=movie_id,
+                                cinema_hall_id=cinema_hall_id)
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
```
Bohdan-Salamakha, line 15: use typehint like: Optional[YOUR_VALUE_TYPE] = None

services/movie_session.py
```diff
@@ -0,0 +1,39 @@
...
+        return MovieSession.objects.filter(show_time__date=session_date)
+    return MovieSession.objects.all()
+
+
+def get_movie_session_by_id(movie_session_id: int) -> QuerySet:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(session_id: int, show_time: datetime = None,
+                         movie_id: int = None, cinema_hall_id: int = None
```
Bohdan-Salamakha, line 26: use typehint like: Optional[YOUR_VALUE_TYPE] = None

format it nicer

Bohdan-Salamakha (changes requested):
v1

services/movie_session.py
```diff
@@ -0,0 +1,47 @@
...
+) -> None:
+    MovieSession.objects.create(show_time=movie_show_time,
+                                movie_id=movie_id,
+                                cinema_hall_id=cinema_hall_id)
+
+
+def get_movies_sessions(session_date: Optional[str] = None) -> QuerySet:
+    if session_date is not None:
+        return MovieSession.objects.filter(show_time__date=session_date)
+    return MovieSession.objects.all()
```
ArsenPidhoretskyi, line 22: I recommend to do like that, it future it will be easier to support your code
```suggestion
    queryset = MovieSession.objects.all()
    if session_date is not None:
        queryset = queryset.filter(show_time__date=session_date)
    return queryset
```

PR #322, opened 2023-04-20
Solution

Ida83:


services/movie.py
```diff
@@ -0,0 +1,39 @@
...
+        return Movie.objects.all()
+    if actors_ids is not None and genres_ids is not None:
+        return Movie.objects.filter(
+            genres__id__in=genres_ids,
+            actors__id__in=actors_ids
+        )
+    if genres_ids is not None and actors_ids is None:
+        return Movie.objects.filter(genres__id__in=genres_ids)
+    if actors_ids is not None and genres_ids is None:
+        return Movie.objects.filter(actors__id__in=actors_ids)
```
ArsenPidhoretskyi, line 15: it is so hard. Just create some variable queryset and then rewrite this variable and apply filter:
queryset = queryset.filter(.....)
In your function must be only two conditions:
1. When genres_ids is not None
2. When actors_ids is not None

ArsenPidhoretskyi (changes requested):

services/movie.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: list = None,
```
olena-hul, line 20: Add Optional to annotations where there is None as default

olena-hul (changes requested):

services/movie.py
Ida83: list added?

services/movie.py
```diff
@@ -0,0 +1,34 @@
+from db.models import Movie
+
+
+def get_movies(genres_ids: list = None, actors_ids: list = None) -> Movie:
```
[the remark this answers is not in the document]
Ida83: what is right QuerySet?

services/movie.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: list = None,
```
olena-hul, line 21:
```suggestion
        genres_ids: Optional[list] = None,
```

olena-hul: And fix in other places too

olena-hul (changes requested):

olena-hul (approved):

olena-hul:
Good job!

PR #325, opened 2023-04-25
Solution

AlexandrZhezheria:


services/movie.py
```diff
@@ -0,0 +1,39 @@
...
+        case (None, None):
+            return Movie.objects.all()
+        case (_, None):
+            return Movie.objects.filter(genres__id__in=genres_ids).distinct()
+        case (None, _):
+            return Movie.objects.filter(actors__id__in=actors_ids).distinct()
+        case _:
+            return Movie.objects.filter(
+                genres__id__in=genres_ids, actors__id__in=actors_ids
+            ).distinct()
```
ArsenPidhoretskyi, line 19: It is not a clear solution. Simplify this code. You must have only two if statements:
1. When genres_ids is not None
2. When actors_ids is not None

services/movie.py
```diff
@@ -0,0 +1,39 @@
+from django.db.models import QuerySet
+from db.models import Movie
+
+
+def get_movies(
+    genres_ids: list[int] = None,
```
ArsenPidhoretskyi, line 6: It accepts Optional[list[int]] because None as default parameter
Fix in all places(with all types)

ArsenPidhoretskyi (changes requested):

services/movie_session.py
```diff
@@ -0,0 +1,49 @@
...
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(pk=movie_session_id)
+
+
+def update_movie_session(
+    session_id: int,
+    show_time: datetime = None,
+    movie_id: int = None,
+    cinema_hall_id: int = None,
```
ArsenPidhoretskyi, line 34: This params also Optional[]

ArsenPidhoretskyi (approved):
Good job!

PR #333, opened 2023-05-09
develop

IvanGuculyak:


vasylhnatiuk (changes requested):
tests are failed

services/cinema_hall.py
```diff
@@ -0,0 +1,19 @@
...
+
+def get_cinema_halls() -> CinemaHall:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(
+        hall_name: str,
+        hall_rows: int,
+        hall_seats_in_row: int
+) -> None:
```
ArsenPidhoretskyi, line 12: It does not return None
It returns CinemaHall

services/cinema_hall.py
```diff
@@ -0,0 +1,19 @@
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> CinemaHall:
```
ArsenPidhoretskyi, line 4: .all does not return one object

services/movie.py
```diff
@@ -0,0 +1,40 @@
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
```
ArsenPidhoretskyi, line 5: Use Optional[list[int]] because None as default provided
Fix in all places with all types

services/movie.py
```diff
@@ -0,0 +1,40 @@
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
+) -> Movie:
```
ArsenPidhoretskyi, line 7: Wrong return type annotations

services/movie.py
```diff
@@ -0,0 +1,40 @@
...
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
+) -> None:
```
ArsenPidhoretskyi, line 28: Wrong return type annotations

services/movie_session.py
```diff
@@ -0,0 +1,58 @@
...
+        cinema_hall_id=cinema_hall_id,
+        movie_id=movie_id
+    )
+
+    return new_movie_session
+
+
+def get_movies_sessions(
+        session_date: datetime = None
+) -> MovieSession:
```
ArsenPidhoretskyi, line 22: Wrong return type annotations

services/movie_session.py
```diff
@@ -0,0 +1,58 @@
...
+    movie_session = MovieSession.objects.all()
```
ArsenPidhoretskyi, line 23: rename movie_session to movie_sessions

services/movie_session.py
```diff
@@ -0,0 +1,58 @@
...
+            id=session_id
+        ).update(show_time=show_time)
+    if movie_id is not None:
+        MovieSession.objects.filter(
+            id=session_id
+        ).update(movie_id=movie_id)
+    if cinema_hall_id is not None:
+        MovieSession.objects.filter(
+            id=session_id
+        ).update(cinema_hall_id=cinema_hall_id)
```
ArsenPidhoretskyi, line 54: No need to do 3 queries. You can do it by one query or use two queries:
1. Get object then set all new variables
2. Save object

ArsenPidhoretskyi (changes requested):

Grayder0152 (approved):

PR #334, opened 2023-05-09
Developer

vborovets0:


services/movie.py
```diff
@@ -0,0 +1,39 @@
+from django.db.models import QuerySet
+
+import init_django_orm  # noqa:F401
+
+from db.models import Movie
+
+
+def get_movies(genres_ids: list[int] = None,
+               actors_ids: list[int] = None) -> QuerySet[Movie]:
```
Bohdan-Salamakha, line 9: use typehint like: Optional[YOUR_VALUE_TYPE] = None

services/movie.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(movie_title: str,
+                 movie_description: str,
+                 genres_ids: list[int] = None,
+                 actors_ids: list[int] = None
```
Bohdan-Salamakha, line 28: use typehint like: Optional[YOUR_VALUE_TYPE] = None

Bohdan-Salamakha (changes requested):
v1

olena-hul (approved):

olena-hul:
Good job!

PR #341, opened 2023-05-23
Develop

SashaLee:


db/models.py
```diff
@@ -5,7 +5,7 @@ class Genre(models.Model):
     name = models.CharField(max_length=255, unique=True)
 
     def __str__(self) -> str:
-        return self.name
+        return f"{self.name}"
```
olena-hul, line 8: Undo this change

services/movie_session.py
```diff
@@ -0,0 +1,47 @@
...
+    queryset = MovieSession.objects.filter(id=session_id)
+
+    if show_time:
+        queryset.update(show_time=show_time)
+
+    if movie_id:
+        queryset.update(movie_id=movie_id)
+
+    if cinema_hall_id:
+        queryset.update(cinema_hall_id=cinema_hall_id)
```
olena-hul, line 43: You're making 3 db queries. Try to do this in 1 query

olena-hul (changes requested):

db/models.py
```diff
@@ -14,3 +14,38 @@ class Actor(models.Model):
...
+        return f"{self.title}"
+
+
+class CinemaHall(models.Model):
+    name = models.CharField(max_length=255)
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
+
+    def __str__(self) -> str:
+        return f"{self.name}"
```
EdAlekseiev, line 35: You don't need to use f-string here. You can use `return self.name` instead of `return f"{self.name}"`

db/models.py
```diff
@@ -14,3 +14,38 @@ class Actor(models.Model):
...
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="actors")
+    genres = models.ManyToManyField(Genre, related_name="genres")
+
+    def __str__(self) -> str:
+        return f"{self.title}"
```
EdAlekseiev, line 26: You don't need to use f-string here. You can use return `self.title` instead of return `f"{self.title}"`

EdAlekseiev (changes requested):

Bohdan-Salamakha (changes requested):

Bohdan-Salamakha (approved):

PR #342, opened 2023-05-24
Solution

Andrew13X:


services/movie.py
```diff
@@ -0,0 +1,41 @@
+import init_django_orm  # noqa: F401
+
+from db.models import Movie
+from django.db.models import QuerySet
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
ArsenPidhoretskyi, line 9: It accepts Optional[list[int]] because None as default
Fix in all places

services/movie.py
```diff
@@ -0,0 +1,41 @@
...
+    if not genres_ids and not actors_ids:
+        return queryset
+    if genres_ids and not actors_ids:
+        return Movie.objects.filter(genres__id__in=genres_ids)
+    if not genres_ids and actors_ids:
+        return Movie.objects.filter(actors__id__in=actors_ids)
+    if genres_ids and actors_ids:
+        queryset = queryset.filter(genres__id__in=genres_ids)
+        queryset = queryset.filter(actors__id__in=actors_ids)
+        return queryset
```
ArsenPidhoretskyi, line 21: So hard code....
You can only two cases:
1.  genres_ids is not None
2. actors_ids is not None

services/movie_session.py
```diff
@@ -0,0 +1,51 @@
...
+        show_time=movie_show_time,
+        cinema_hall=CinemaHall.objects.get(id=cinema_hall_id),
+        movie=Movie.objects.get(id=movie_id)
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
+    if session_date:
+        return MovieSession.objects.filter(show_time__date=session_date)
+    return MovieSession.objects.all()
```
ArsenPidhoretskyi, line 23:
```suggestion
    movie_sessions = MovieSession.objects.all()
    if session_date:
        movie_sessions = movie_sessions.filter(show_time__date=session_date)
    return movie_sessions

```
This code easiest to support in future

services/movie_session.py
```diff
@@ -0,0 +1,51 @@
...
+            show_time=show_time,
+        )
+    if movie_id:
+        MovieSession.objects.filter(id=session_id).update(
+            movie=movie_id,
+        )
+    if cinema_hall_id:
+        MovieSession.objects.filter(id=session_id).update(
+            cinema_hall=cinema_hall_id,
+        )
```
ArsenPidhoretskyi, line 47: No need to do 3 update statement do it only in 2 or 1 queries. Two possible options:
1. Get an object, set all fields, and save
2. Create some dict which field we need to update and then do update

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #351, opened 2023-05-31
Solution

Fawnygod:


YukihiroSM (approved):
Looks good! 

Also, I would like to recommend trying to use black formatter to fix formatting issues, like `'` instead of `"` in migrations :)

PR #353, opened 2023-06-02
Solution

Dimdimior:


services/movie.py
```diff
@@ -0,0 +1,35 @@
...
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(movie_title: str,
+                 movie_description: str,
+                 genres_ids: list[int] = None,
+                 actors_ids: list[int] = None
+                 ) -> Movie:
```
Bohdan-Salamakha, line 25:
```suggestion
def create_movie(
                 movie_title: str,
                 movie_description: str,
                 genres_ids: list[int] = None,
                 actors_ids: list[int] = None
                 ) -> Movie:
```

Bohdan-Salamakha: use typehint like: Optional[YOUR_VALUE_TYPE] = None

Bohdan-Salamakha (changes requested):

YukihiroSM (approved):
The code is good!
In future I recommend you to use `related_name` within relations

PR #354, opened 2023-06-03
Solution

vlamuha:


services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+    queryset = MovieSession.objects.filter(id=session_id)
+
+    if show_time is not None:
+        queryset.update(show_time=show_time)
+
+    if movie_id is not None:
+        queryset.update(movie_id=movie_id)
+
+    if cinema_hall_id is not None:
+        queryset.update(cinema_hall_id=cinema_hall_id)
```
ArsenPidhoretskyi, line 44: You do 3 queries. It is not okay. You have 2 options:
1. Get the object, set all new values, and then save it(2 queries)
2. Create and form a dictionary with fields to update and then do .update(1 query)

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):
Well done!

PR #365, opened 2023-06-16
Solution

gleblitvinenko:


services/movie_session.py
```diff
@@ -0,0 +1,62 @@
...
+    updated_movie = MovieSession.objects.filter(id=session_id)
+
+    if show_time:
+        updated_movie.update(show_time=show_time)
+
+    if movie_id:
+        updated_movie.update(movie=movie_id)
+
+    if cinema_hall_id:
+        updated_movie.update(cinema_hall_id=cinema_hall_id)
```
ArsenPidhoretskyi, line 54: No need to do 3 queriest. You can do it in two queries(get, set all values and save object) or 1 query(create dict and do update once)

ArsenPidhoretskyi (changes requested):

db/migrations/0002_cinemahall_movie_moviesession.py
```diff
@@ -0,0 +1,42 @@
+# Generated by Django 4.0.2 on 2023-06-16 19:50
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('db', '0001_initial'),
```
vasylhnatiuk, line 10: change ' to ", use Black

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #369, opened 2023-06-27
Solution

dmitriy-kds:


services/movie.py
```diff
@@ -0,0 +1,40 @@
+from typing import List
+from django.db.models import QuerySet
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: List[int] = None,
+        actors_ids: List[int] = None
```
EdAlekseiev, line 8: If the variable can be None you must show this fact in type-hint with the help `Optional[your variable type]` or 'your variable type | None'. 
For example: `genres_ids: Optional[List[int]] = None` or `genres_ids: List[int] | None = None`

services/movie.py
```diff
@@ -0,0 +1,40 @@
...
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: List[int] = None,
+        actors_ids: List[int] = None
```
EdAlekseiev, line 29: The same comment as above

services/movie_session.py
```diff
@@ -0,0 +1,51 @@
...
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: str = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
```
EdAlekseiev, line 34: The same comment as above

EdAlekseiev (changes requested):

Bohdan-Salamakha (approved):

PR #372, opened 2023-06-30
Solution

diana-shyrokikh:


db/models.py
```diff
@@ -14,3 +14,43 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="movies")
+    genres = models.ManyToManyField(Genre, related_name="movies")
```
volodymyr-komarnyckyi, line 23: The related_names can not be the same, since this would introduce confusion

db/models.py
```diff
@@ -14,3 +14,43 @@ class Actor(models.Model):
...
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(
+        CinemaHall,
+        related_name="movie_sessions",
+        on_delete=models.CASCADE
+    )
+    movie = models.ForeignKey(
+        Movie,
+        related_name="movie_sessions",
```
volodymyr-komarnyckyi, line 51: Same here

volodymyr-komarnyckyi (changes requested):

volodymyr-komarnyckyi (approved):
Good job))

db/models.py
Oleg-s-s: why not?

db/models.py
Oleg-s-s: why?

db/models.py
Oleg-s-s, line 1: previous version was ok

services/movie.py
```diff
@@ -0,0 +1,42 @@
...
+        description=movie_description,
+    )
+
+    if genres_ids:
+        new_movie.genres.set(genres_ids)
+
+    if actors_ids:
+        new_movie.actors.set(actors_ids)
+
+    return new_movie
```
Oleg-s-s, line 42: same here

Oleg-s-s (changes requested):

Oleg-s-s (approved):
Good job

PR #375, opened 2023-07-02
'Solution'

ant-komarov:


imelnyk007 (approved):

services/movie_session.py
```diff
@@ -0,0 +1,52 @@
...
+
+    if movie_id:
+        session_to_update.update(movie_id=movie_id)
+
+    if cinema_hall_id:
+        session_to_update.update(cinema_hall_id=cinema_hall_id)
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    MovieSession.objects.filter(id=session_id).delete()
```
Oleg-s-s, line 52: You can use `some_name = Model.objects.filter().first()` syntax to avoid calling `try, except` if you are not sure if entry exist. And then just `if some_name: ...`. But in your case (not providing any checks)  looks like it would be better to use `get()` instead

Oleg-s-s (changes requested):

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
...
+
+class CinemaHall(models.Model):
+    name = models.CharField(max_length=255)
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
+
+    def __str__(self) -> str:
+        return self.name
+
+    @property
```
LiudmylaKulinchenko, line 37: you can remove decorator

services/movie.py
```diff
@@ -0,0 +1,42 @@
...
+        description=movie_description
+    )
+
+    if genres_ids:
+        new_movie.genres.set(genres_ids)
+
+    if actors_ids:
+        new_movie.actors.set(actors_ids)
+
+    return new_movie
```
LiudmylaKulinchenko, line 42: No need to return anything here

services/movie_session.py
```diff
@@ -0,0 +1,52 @@
...
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: datetime = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> None:
+    session_to_update = MovieSession.objects.filter(id=session_id)
```
LiudmylaKulinchenko, line 39: Use `.get()` here to get single object

LiudmylaKulinchenko (changes requested):

db/models.py
ant-komarov: 3 tests crash if I remove decorator

LiudmylaKulinchenko: Yes, it is a better practice in this case, so you should keep it

LiudmylaKulinchenko (approved):
Good job! 🐣🌼☀

PR #378, opened 2023-07-02
Solution

Po-SvinSkipodla:


db/models.py
```diff
@@ -14,3 +14,43 @@ class Actor(models.Model):
...
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(
+        CinemaHall,
+        related_name="cinema_halls",
+        on_delete=models.CASCADE
+    )
+    movie = models.ForeignKey(
+        Movie,
+        related_name="movies",
```
YevheniiIovchu, line 51: It's bad practice to use the same related name as a column name. And in all cases above. Don't forget about making and saving migration before you push

YevheniiIovchu (changes requested):

db/models.py
Oleg-s-s, line 51: fix related_name, look in the checklist

Oleg-s-s (changes requested):

LiudmylaKulinchenko (approved):
Good job 🌺🌼

PR #383, opened 2023-07-03
'Solution'

RedCatHot:


YKonoplyov (changes requested):
Don`t forget to add migrations to git

services/movie_session.py
```diff
@@ -0,0 +1,49 @@
...
+) -> None:
+    movie_session = get_movie_session_by_id(session_id)
+    if show_time:
+        movie_session.show_time = show_time
+    if movie_id:
+        movie = get_movie_by_id(movie_id)
+        movie_session.movie = movie
+    if cinema_hall_id:
+        cinema_hall = CinemaHall.objects.get(id=cinema_hall_id)
+        movie_session.cinema_hall = cinema_hall
```
YuriiKindrat, line 44: could u use lookups insted ? 
mvie_session.movie_id = movie_id

YuriiKindrat (changes requested):

services/cinema_hall.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+def get_cinema_halls() -> QuerySet:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(
+    hall_name: str, hall_rows: int, hall_seats_in_row: int
+) -> CinemaHall:
+    return CinemaHall.objects.create(
```
LiudmylaKulinchenko, line 13: No need to use return in create/update/delete methods

LiudmylaKulinchenko (changes requested):

PR #385, opened 2023-07-03
Solution

YevheniiIovchu:


DmytroTsehelnyk (changes requested):

services/movie.py
```diff
@@ -0,0 +1,39 @@
...
+    if genres_ids and actors_ids:
+        return Movie.objects.filter(
+            genres__id__in=genres_ids
+        ).filter(
+            actors__id__in=actors_ids
+        ).distinct()
+    if genres_ids:
+        return Movie.objects.filter(genres__id__in=genres_ids).distinct()
+    if actors_ids:
+        return Movie.objects.filter(actors__id__in=actors_ids).distinct()
```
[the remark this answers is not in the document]
YevheniiIovchu: If we received as input in actors_ids the different actors that acted in the same movie, we would get the same entry in our table without distinct.

db/models.py
```diff
@@ -14,3 +14,49 @@ class Actor(models.Model):
...
+
+class CinemaHall(models.Model):
+    name = models.CharField(max_length=255)
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
+
+    def __str__(self) -> str:
+        return self.name
+
+    @property
```
LiudmylaKulinchenko, line 43: using the decorator is not necessary. You can remove it

db/models.py
```diff
@@ -14,3 +14,49 @@ class Actor(models.Model):
...
+    def capacity(self) -> int:
+        return self.rows * self.seats_in_row
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(
+        CinemaHall,
+        related_name="sessions_in_cinema_hall",
```
LiudmylaKulinchenko, line 52: Is there a need in such detalized related_name? Use advice from the checklist (Code style, second point)

services/movie.py
```diff
@@ -0,0 +1,31 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(genres_ids: list[int] = None,
+               actors_ids: list[int] = None) -> QuerySet[Movie]:
```
LiudmylaKulinchenko, line 7: I like that you specified `QuerySet[Movie]` here

LiudmylaKulinchenko (changes requested):

db/models.py
YevheniiIovchu: But Danylo said in video Calculated Field of this topic, that using decorator property in such cases is an example of good code style and without this decorator 3 tests fail.

LiudmylaKulinchenko: Yes, it is, you are right

db/models.py
```diff
@@ -14,3 +14,49 @@ class Actor(models.Model):
...
+    @property
+    def capacity(self) -> int:
+        return self.rows * self.seats_in_row
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(
+        CinemaHall,
+        related_name="sessions",
```
LiudmylaKulinchenko, line 52: probably, movie_sessions would be better, just like model name, but it is fine! Just my opinion

LiudmylaKulinchenko (approved):
Good job 🔥

PR #388, opened 2023-07-03
Solution

Gvilion:


settings.py
```diff
@@ -21,6 +21,6 @@
 
 USE_I18N = True
 
-USE_TZ = False
+USE_TZ = True
```
Po-SvinSkipodla, line 24: U can set USE_TZ back to False, and ur tests will pass without warnings

Po-SvinSkipodla (approved):

services/cinema_hall.py
```diff
@@ -0,0 +1,19 @@
...
+
+def get_cinema_halls() -> QuerySet:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(
+        hall_name: str,
+        hall_rows: int,
+        hall_seats_in_row: int
+) -> QuerySet:
```
YuriiKindrat, line 14: are u sure about return type ?

services/movie_session.py
```diff
@@ -0,0 +1,49 @@
...
+        session.show_time = show_time
+    if movie_id:
+        session.movie_id = movie_id
+    if cinema_hall_id:
+        session.cinema_hall_id = cinema_hall_id
+    session.save()
+    return session
+
+
+def delete_movie_session_by_id(session_id: int) -> MovieSession:
```
YuriiKindrat, line 48: are u sure the delete method returns smth ?

YuriiKindrat (changes requested):

Oleg-s-s (approved):
:+1:

PR #391, opened 2023-07-03
Solution

Kasatello:


IhorVoskoboinikov (approved):
Great work!!! There are some comment to your code:

1. I think we may use "relation name" in 
actors = models.ManyToManyField(Actor)
genres = models.ManyToManyField(Genre)

2. I think better use
cinema_hall_id=cinema_hall_id
movie_id=movie_id
it's the same but better read

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
```
LiudmylaKulinchenko, line 23: Add related_name

LiudmylaKulinchenko (changes requested):

services/cinema_hall.py
```diff
@@ -0,0 +1,15 @@
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> str:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(hall_name: str,
+                       hall_rows: int,
+                       hall_seats_in_row: int) -> str:
```
YuriiKindrat, line 10: same,fix this everywhere

YuriiKindrat (changes requested):

db/models.py
```diff
@@ -19,8 +19,8 @@ def __str__(self) -> str:
 class Movie(models.Model):
     title = models.CharField(max_length=255)
     description = models.TextField()
-    actors = models.ManyToManyField(Actor, related_name="movie_actors")
-    genres = models.ManyToManyField(Genre, related_name="movie_genres")
+    actors = models.ManyToManyField(Actor, related_name="actors_of_movies")
+    genres = models.ManyToManyField(Genre, related_name="genres_of_movies")
```
YuriiKindrat, line 23: imagine how would u call it using orm 
`Actor.objects.get(<some_id>).actors_of_movies` to get all related movies to an actor 
maybe it is better to use `movies` here ?

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
...
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="actors_of_movies")
+    genres = models.ManyToManyField(Genre, related_name="genres_of_movies")
+
+    def __str__(self) -> str:
+        return f"{self.title}"
```
YuriiKindrat, line 26: why do u need f-string here?

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
...
+
+
+class CinemaHall(models.Model):
+    name = models.CharField(max_length=255)
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
+
+    def __str__(self) -> str:
+        return f"{self.name}"
```
YuriiKindrat, line 35: same

services/cinema_hall.py
```diff
@@ -0,0 +1,17 @@
...
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> QuerySet:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(hall_name: str,
+                       hall_rows: int,
+                       hall_seats_in_row: int) -> QuerySet:
```
YuriiKindrat, line 12: are u usre that create method return queryset ?

YuriiKindrat: fix this everywhere

services/movie_session.py
```diff
@@ -0,0 +1,42 @@
+from django.db.models import QuerySet
+
+from db.models import MovieSession, CinemaHall, Movie
+
+
+def create_movie_session(movie_show_time: str,
+                         cinema_hall_id: int,
+                         movie_id: int) -> str:
```
YuriiKindrat, line 8: are u sure about str ?

YuriiKindrat (changes requested):

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
...
+        return str(self.title)
+
+
+class CinemaHall(models.Model):
+    name = models.CharField(max_length=255)
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
+
+    def __str__(self) -> str:
+        return str(self.name)
```
YuriiKindrat, line 35: why do u need str func here ?

services/movie.py
```diff
@@ -0,0 +1,34 @@
...
+        return movie.filter(genres__id__in=genres_ids,
+                            actors__id__in=actors_ids)
+    if genres_ids:
+        return movie.filter(genres__id__in=genres_ids)
+    if actors_ids:
+        return movie.filter(actors__id__in=actors_ids)
+    return movie
+
+
+def get_movie_by_id(movie_id: int) -> QuerySet:
```
YuriiKindrat, line 19: does get method returns queryset ?

services/movie_session.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
+    session = MovieSession.objects.all()
+    if session_date:
+        return session.filter(show_time__date=session_date)
+    return session
+
+
+def get_movie_session_by_id(movie_session_id: int) -> QuerySet:
```
YuriiKindrat, line 23: same

YuriiKindrat (changes requested):

Oleg-s-s (approved):
:+1:

PR #393, opened 2023-07-03
Movie, CinemaHall, and MovieSession:

IhorVoskoboinikov:
(first_commit)


Kasatello (approved):
Well done 😎

LiudmylaKulinchenko (changes requested):

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField("Actor", related_name="movie_actors")
+    genres = models.ManyToManyField("Genre", related_name="movie_genres")
```
Oleg-s-s, line 23: Nor critical but from checklist:
Use the correct `related_name`. In the majority of cases, the name of the model in plural will be just right.

services/movie.py
```diff
@@ -0,0 +1,47 @@
...
+    elif genres_ids and actors_ids is None:
+        queryset = queryset.filter(genres__id__in=genres_ids)
+    elif genres_ids is None and actors_ids:
+        queryset = queryset.filter(actors__id__in=actors_ids)
+
+    return queryset
+
+
+def get_movie_by_id(movie_id: int) -> QuerySet:
+    return Movie.objects.get(id=movie_id)
```
Oleg-s-s, line 29: Are you sure this return QuerySet?

Oleg-s-s (changes requested):
Good,

YuriiKindrat (approved):

PR #396, opened 2023-07-04
my_commit

Mr-Furman:


services/movie_session.py
```diff
@@ -0,0 +1,45 @@
...
+
+def update_movie_session(session_id: int,
+                         show_time: datetime = None,
+                         movie_id: int = None,
+                         cinema_hall_id: int = None) -> None:
+    movie_session = MovieSession.objects.get(id=session_id)
+    if show_time:
+        movie_session.show_time = show_time
+    if movie_id:
+        movie = Movie.objects.get(id=movie_id)
```
OlhaTryhub, line 36: You dont really need import Movie class and use it to change movie_id for MovieSession. You have movie_session instance and you can set id by next way:

services/movie_session.py
```diff
@@ -0,0 +1,45 @@
...
+        movie_session.movie = movie
+    if cinema_hall_id:
+        hall = CinemaHall.objects.get(id=cinema_hall_id)
+        movie_session.cinema_hall = hall
```
OlhaTryhub, line 40: The same as with movie_id:
movie_session.cinema_hall_id = cinema_hall_id

OlhaTryhub (changes requested):

services/movie_session.py
```diff
@@ -0,0 +1,45 @@
...
+def get_movies_sessions(session_date: str = None) -> QuerySet:
+    query = MovieSession.objects.all()
+
+    if session_date:
+        query = MovieSession.objects.filter(show_time__date=session_date)
+
+    return query
+
+
+def get_movie_session_by_id(movie_session_id: int) -> QuerySet:
```
YuriiKindrat, line 24: are u sure that get method returns QuerySet

YuriiKindrat (changes requested):

Oleg-s-s (approved):
:+1:

PR #400, opened 2023-07-06
Solution

WalDragnir:


services/movie.py
```diff
@@ -0,0 +1,33 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(genres_ids: list[int] = None,
```
ArsenPidhoretskyi, line 6: It accepts Optional[list[int]] because None as default. Fix in all palces with all types

ArsenPidhoretskyi (changes requested):

services/movie.py
WalDragnir: image
With Optional test issues

db/models.py
```diff
@@ -14,3 +14,44 @@ class Actor(models.Model):
...
+    )
+    movie = models.ForeignKey(
+        Movie,
+        on_delete=models.CASCADE,
+        related_name="movie_sessions"
+    )
+
+    def __str__(self) -> str:
+        return f"{self.movie.title} " \
+               f"{self.show_time.strftime('%Y-%m-%d %H:%M:%S')}"
```
Grayder0152, line 57: use `()` instead of `\`

services/movie.py
```diff
@@ -0,0 +1,35 @@
...
+
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(genres_ids: Optional[list[int]] = None,
+               actors_ids: Optional[list[int]] = None) -> QuerySet:
+    queryset = Movie.objects.all()
+    if genres_ids:
```
Grayder0152, line 11:
```suggestion
    if genres_ids is not None:

```

services/movie_session.py
```diff
@@ -0,0 +1,48 @@
...
+
+    return queryset
+
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(session_id: int,
+                         show_time: str = None,
```
Grayder0152, line 30:
```suggestion
                         show_time: Optional[str] = None,

```

Grayder0152 (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #406, opened 2023-07-09
Solution

NazarLytvyniuk:


services/movie.py
```diff
@@ -0,0 +1,44 @@
+from django.db.models.query import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
Bohdan-Salamakha, line 8: use typehint like: Optional[YOUR_VALUE_TYPE] = None

services/movie.py
```diff
@@ -0,0 +1,44 @@
...
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
Bohdan-Salamakha, line 31: use typehint like: Optional[YOUR_VALUE_TYPE] = None

services/movie_session.py
```diff
@@ -0,0 +1,58 @@
...
+        cinema_hall_id: int
+) -> MovieSession:
+    return MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
```
Bohdan-Salamakha, line 19: use typehint like: Optional[YOUR_VALUE_TYPE] = None

services/movie_session.py
```diff
@@ -0,0 +1,58 @@
...
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: datetime = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
```
Bohdan-Salamakha, line 41: use typehint like: Optional[YOUR_VALUE_TYPE] = None

Bohdan-Salamakha (changes requested):

db/models.py
```diff
@@ -14,3 +14,43 @@ class Actor(models.Model):
...
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(
+        CinemaHall,
+        on_delete=models.CASCADE,
+        related_name="movies_sessions"
+    )
+    movie = models.ForeignKey(
+        Movie,
+        on_delete=models.CASCADE,
+        related_name="movies_sessions"
```
Dimosphen1, line 52: A related name could be another here, because calling for MovieSessions from Movie would require duplication in naming, e.g. `Movie.movies_sessions.filter(...`, when it is already clear that sessions relate to the movie. You could use just `sessions` here.

db/models.py
```diff
@@ -14,3 +14,43 @@ class Actor(models.Model):
...
+    def __str__(self) -> str:
+        return self.name
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(
+        CinemaHall,
+        on_delete=models.CASCADE,
+        related_name="movies_sessions"
```
Dimosphen1, line 47: `movie_sessions` is better than `movies_sessions`, because there will be one movie in a session when referring to `MovieSession` records from `CinemaHall`.

Dimosphen1 (approved):
LGTM, several minor comments left regarding naming. Nice work!

PR #407, opened 2023-07-11
Solution

arturiermolenko:


services/movie.py
```diff
@@ -0,0 +1,39 @@
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
Bohdan-Salamakha, line 6: use typehint like: Optional[YOUR_VALUE_TYPE] = None

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+    movie_session = MovieSession(
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
+    movie_session.save()
+    return movie_session
+
+
+def get_movies_sessions(session_date: str = None) -> MovieSession:
```
Bohdan-Salamakha, line 18: use typehint like: Optional[YOUR_VALUE_TYPE] = None

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+
+def get_movie_session_by_id(id_: int) -> MovieSession:
+    return MovieSession.objects.get(id=id_)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: str = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
```
Bohdan-Salamakha, line 35: use typehint like: Optional[YOUR_VALUE_TYPE] = None

Bohdan-Salamakha (changes requested):

Bohdan-Salamakha (approved):

PR #408, opened 2023-07-11
Initial implementation

thorium-zero:


db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
...
+        return self.rows * self.seats_in_row
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey("CinemaHall", on_delete=models.CASCADE)
+    movie = models.ForeignKey(Movie, on_delete=models.CASCADE)
+
+    def __str__(self) -> str:
+        return f"{self.movie} {self.show_time.strftime('%Y-%m-%d %H:%M:%S')}"
```
YuriiKindrat, line 48: show time will be saved at the same format, so there is no need to use strftime )

services/cinema_hall.py
```diff
@@ -0,0 +1,17 @@
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> list[CinemaHall]:
+    return CinemaHall.objects.all()
```
YuriiKindrat, line 5: are u sure about list type annotation ? 
it seems to me, that method all returns queryset)

services/movie.py
```diff
@@ -0,0 +1,38 @@
+from db.models import Movie
+from django.db.models import Q
+
+
+def get_movies(
+    genres_ids: list[int] = None,
+    actors_ids: list[int] = None
+) -> list[Movie]:
```
YuriiKindrat, line 8: same is here)

YuriiKindrat (changes requested):

services/cinema_hall.py
thorium-zero: ffs, I knew I'm going to forget to change this :(

services/movie_session.py
```diff
@@ -0,0 +1,55 @@
...
+from db.models import CinemaHall, Movie, MovieSession
+
+
+def create_movie_session(
+    movie_show_time: datetime,
+    movie_id: int,
+    cinema_hall_id: int
+) -> MovieSession | None:
+    movie = Movie.objects.get(id=movie_id)
+    cinema_hall = CinemaHall.objects.get(id=cinema_hall_id)
```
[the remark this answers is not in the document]
thorium-zero: yeah, I was going to add some `nullability` checks here, so I could validate incoming parameters, but then I forgot to do that. So yeah, those are useless here

db/models.py
thorium-zero: oh, ok, I didn't realize that, thanks

YuriiKindrat (approved):

PR #415, opened 2023-07-21
Solution_1

CeO-PsH:


services/movie_session.py
```diff
@@ -0,0 +1,62 @@
...
+        cinema_hall_id=cinema_hall_id,
+        movie_id=movie_id
+    )
+
+
+def get_movies_sessions(
+        session_date: str = None
+) -> QuerySet:
+    if session_date:
+        return MovieSession.objects.all().filter(
```
Polyakiv-Andrey, line 23: If you use a filter() then all() is not necessary

services/movie_session.py
```diff
@@ -0,0 +1,62 @@
...
+            show_time__date=session_date
+        )
+    else:
```
Polyakiv-Andrey, line 26: else should be omitted

Polyakiv-Andrey (changes requested):

services/movie_session.py
```diff
@@ -0,0 +1,61 @@
...
+        cinema_hall_id: int = None
+) -> QuerySet:
+    sess = MovieSession.objects.all()
+
+    if show_time is not None:
+        sess.update(
+            show_time=show_time
+        )
+    if movie_id is not None:
+        sess.filter(id=session_id).update(
```
pavlejviki, line 46: you can simplify your code by retrieving the needed movie session object using .get() method before all the **if** statements and than assign to its attributes the needed values :

and after all the **if** statements you can save all the changes using .save() method.

services/movie_session.py
```diff
@@ -0,0 +1,61 @@
...
+        sess.update(
+            cinema_hall=cinema_hall_id
+        )
+    return sess
+
+
+def delete_movie_session_by_id(
+        session_id: int
+) -> MovieSession:
+    return MovieSession.objects.filter(
```
pavlejviki, line 59: its not necessary to use **return** here.

pavlejviki (changes requested):

services/movie_session.py
CeO-PsH: Thank you))))))

Polyakiv-Andrey (approved):
Good job!

PR #416, opened 2023-07-22
Solution

vkrasnovyd:


services/movie.py
```diff
@@ -0,0 +1,44 @@
+from django.db.models import QuerySet
+
+import init_django_orm  # noqa: F401
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
olena-hul, line 10: Add Optional to the annotation where the default argument can be None

olena-hul: Fix everywhere

services/movie_session.py
```diff
@@ -0,0 +1,59 @@
...
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
+    queryset = MovieSession.objects.all()
+
+    if session_date:
+        year_, month_, day_ = session_date.split("-")
+        queryset = queryset.filter(show_time__year=year_).filter(
+            show_time__month=month_).filter(show_time__day=day_)
```
olena-hul, line 29: You can filter directly by date, use `strptime`

olena-hul (changes requested):

olena-hul (approved):

olena-hul:
Good job!

PR #417, opened 2023-07-24
'Solution'

StinGER-ua:


services/movie.py
```diff
@@ -0,0 +1,40 @@
+from db.models import Movie
+from db.models import models
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
ArsenPidhoretskyi, line 7: It accepts Optional[list[int]] due to default None
Fix in all places

services/movie_session.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def create_movie_session(
+        movie_show_time: datetime.datetime,
+        movie_id: int,
+        cinema_hall_id: int
+) -> None:
+
+    movie_ = Movie.objects.get(id=movie_id)
+    hall = CinemaHall.objects.get(id=cinema_hall_id)
```
ArsenPidhoretskyi, line 14: No need to use get you can do: fk_id

services/movie.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: Optional[int] = None,
```
ArsenPidhoretskyi, line 29:
```suggestion
        genres_ids: Optional[list[int]] = None,

```
the same with actors

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #422, opened 2023-08-01
Done!

DenysRaboshchuk:


services/movie.py
```diff
@@ -0,0 +1,40 @@
+from db.models import Movie
+from django.db.models import QuerySet
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```

services/movie.py
```diff
@@ -0,0 +1,40 @@
...
+    queryset = Movie.objects.all()
+    if genres_ids is not None:
+        queryset = queryset.filter(genres__id__in=genres_ids)
+    if actors_ids is not None:
+        queryset = queryset.filter(actors__id__in=actors_ids)
+
+    return queryset
+
+
+def get_movie_by_id(movie_id: int) -> QuerySet:
```
Bohdan-Salamakha, line 19: invalid return typehint

services/movie_session.py
```diff
@@ -0,0 +1,55 @@
...
+    new_movie_session = MovieSession.objects.create(
+        show_time=movie_show_time,
+        cinema_hall_id=cinema_hall_id,
+        movie_id=movie_id
+    )
+
+    return new_movie_session
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
```
Bohdan-Salamakha, line 22: use typehint like: Optional[YOUR_VALUE_TYPE] = None

Bohdan-Salamakha (changes requested):

Bohdan-Salamakha (approved):

PR #434, opened 2023-08-19
Solution

romkapomka12:


services/movie.py
```diff
@@ -0,0 +1,44 @@
+from db.models import Movie, Actor, Genre
+from typing import List
+
+
+def get_movies(
+        genres_ids: List[int] = None,
+        actors_ids: List[int] = None
```

services/movie.py
```diff
@@ -0,0 +1,44 @@
...
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: List[int] = None,
+        actors_ids: List[int] = None) -> Movie:
```
Bohdan-Salamakha, line 32: use typehint like: Optional[YOUR_VALUE_TYPE] = None
+

```suggestion
def create_movie(
        movie_title: str,
        movie_description: str,
        genres_ids: List[int] = None,
        actors_ids: List[int] = None
        ) -> Movie:

```

services/movie_session.py
```diff
@@ -0,0 +1,43 @@
...
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id)
+
+
+def get_movies_sessions(session_date: str = None) -> MovieSession:
+    if session_date:
+        date_session = datetime.strptime(session_date, "%Y-%m-%d")
+        return MovieSession.objects.filter(show_time__date=date_session)
+    else:
```
Bohdan-Salamakha, line 18: remove redundant "else"

services/movie_session.py
```diff
@@ -0,0 +1,43 @@
...
+def create_movie_session(movie_show_time: date,
+                         movie_id: int,
+                         cinema_hall_id: int) -> MovieSession:
+    return MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id)
+
+
+def get_movies_sessions(session_date: str = None) -> MovieSession:
```
Bohdan-Salamakha, line 14: use typehint like: Optional[YOUR_VALUE_TYPE] = None
+
invalid return typehint

services/movie_session.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(session_id: int,
+                         show_time: str = None,
+                         movie_id: int = None,
+                         cinema_hall_id: int = None) -> MovieSession:
```
Bohdan-Salamakha, line 29: use typehint like: Optional[YOUR_VALUE_TYPE] = None

Bohdan-Salamakha (changes requested):

olena-hul (changes requested):

services/movie_session.py
```diff
@@ -0,0 +1,43 @@
...
+def create_movie_session(movie_show_time: date,
+                         movie_id: int,
+                         cinema_hall_id: int) -> MovieSession:
+    return MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id)
+
+
+def get_movies_sessions(session_date: Optional[str] = None) -> MovieSession:
```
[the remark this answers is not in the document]
romkapomka12: because  QuerySet[MovieSession]?

db/models.py
```diff
@@ -14,3 +14,38 @@ class Actor(models.Model):
...
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
+
+    @property
+    def capacity(self) -> int:
+        capacity = self.rows * self.seats_in_row
+        return capacity
+
+    def __str__(self) -> str:
+        return f"{self.name}"
```
EdAlekseiev, line 41: `__str__` method must be above `capacity` method

EdAlekseiev (changes requested):

db/models.py
romkapomka12: Because magical methods always go first?

PR #440, opened 2023-08-28
Many to many

IProskurnytskyi:


services/movie.py
```diff
@@ -0,0 +1,38 @@
+import init_django_orm  # noqa: F401
+
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
Bohdan-Salamakha, line 10: use typehint like: Optional[YOUR_VALUE_TYPE] = None

services/movie.py
```diff
@@ -0,0 +1,38 @@
...
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
Bohdan-Salamakha, line 29: use typehint like: Optional[YOUR_VALUE_TYPE] = None

services/movie_session.py
```diff
@@ -0,0 +1,48 @@
...
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: str = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
```
Bohdan-Salamakha, line 34: use typehint like: Optional[YOUR_VALUE_TYPE] = None

pavlejviki (approved):
Well done!

PR #441, opened 2023-08-30
Implemented Movie-Moviesession-CinemaHall task

OleksiiKiva:


services/movie.py
```diff
@@ -0,0 +1,40 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
Bohdan-Salamakha, line 8: use typehint like: Optional[YOUR_VALUE_TYPE] = None

services/movie.py
```diff
@@ -0,0 +1,40 @@
...
+    if genres_ids:
+        searched_movie = searched_movie.filter(genres__id__in=genres_ids)
+
+    if actors_ids:
+        searched_movie = searched_movie.filter(actors__id__in=actors_ids)
+
+    return searched_movie
+
+
+def get_movie_by_id(movie_id: int) -> QuerySet:
```
Bohdan-Salamakha, line 21: invalid return typehint

services/movie.py
```diff
@@ -0,0 +1,40 @@
...
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
Bohdan-Salamakha, line 29: use typehint like: Optional[YOUR_VALUE_TYPE] = None

services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+        cinema_hall_id: int
+) -> None:
+    MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
+
+
+def get_movies_sessions(session_date: datetime = None) -> QuerySet:
```
Bohdan-Salamakha, line 20: use typehint like: Optional[YOUR_VALUE_TYPE] = None

services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+
+    if session_date:
+        searched_session = searched_session.filter(
+            show_time__date=session_date
+        )
+
+    return searched_session
+
+
+def get_movie_session_by_id(movie_session_id: int) -> QuerySet:
```
Bohdan-Salamakha, line 31: invalid return typehint

services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: datetime = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
```
Bohdan-Salamakha, line 39: use typehint like: Optional[YOUR_VALUE_TYPE] = None

Bohdan-Salamakha (changes requested):

services/movie.py
OleksiiKiva: Bogdan, thanks for your help. I carefully read about Optional[...]

services/movie.py
OleksiiKiva: Yes, it's my carelessness

services/movie_session.py
```diff
@@ -0,0 +1,55 @@
...
+    updated_session = MovieSession.objects.filter(id=session_id)
+
+    if show_time:
+        updated_session.update(show_time=show_time)
+
+    if movie_id:
+        updated_session.update(movie_id=movie_id)
+
+    if cinema_hall_id:
+        updated_session.update(cinema_hall_id=cinema_hall_id)
```
olena-hul, line 51: Use 1 query to update the records, now you're using 3

olena-hul (changes requested):

services/movie_session.py
OleksiiKiva: Wow, really. Thanks for the optimization. I rewrote this part

YukihiroSM (approved):
Good job! Thanks for fixing this!

PR #442, opened 2023-08-30
Solution

V-Shkrobatskyi:


db/models.py
```diff
@@ -14,3 +14,43 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="actors")
+    genres = models.ManyToManyField(Genre, related_name="genres")
```
ArsenPidhoretskyi, line 23: Wrong related names

db/models.py
```diff
@@ -14,3 +14,43 @@ class Actor(models.Model):
...
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(
+        CinemaHall,
+        on_delete=models.DO_NOTHING,
+        related_name="movie_sessions"
+    )
+    movie = models.ForeignKey(
+        Movie,
+        on_delete=models.DO_NOTHING,
+        related_name="movies"
```
ArsenPidhoretskyi, line 52: Wrong related name

services/cinema_hall.py
```diff
@@ -0,0 +1,17 @@
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> list[CinemaHall]:
```
ArsenPidhoretskyi, line 4: .all() does not return list it returns QuerySet

services/movie.py
```diff
@@ -0,0 +1,39 @@
...
+    queryset = Movie.objects.all()
+    if None not in (genres_ids, actors_ids):
+        return queryset.filter(
+            actors__id__in=actors_ids,
+            genres__id__in=genres_ids,
+        )
+    elif genres_ids is not None:
+        return queryset.filter(genres__id__in=genres_ids)
+    elif actors_ids is not None:
+        return queryset.filter(actors__id__in=actors_ids)
```
ArsenPidhoretskyi, line 17: So hard code. You just have two cases:
1. if genres_ids is not None
2. if actors_ids is not None

services/movie.py
```diff
@@ -0,0 +1,39 @@
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None,
+) -> list[Movie]:
```
ArsenPidhoretskyi, line 7: Wrong annotations

services/movie.py
```diff
@@ -0,0 +1,39 @@
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
```
ArsenPidhoretskyi, line 5: It accepts Optional[list[int]] because None as default. Fix in all places

services/movie_session.py
```diff
@@ -0,0 +1,43 @@
...
+        show_time=movie_show_time,
+        movie=Movie.objects.get(id=movie_id),
+        cinema_hall=CinemaHall.objects.get(id=cinema_hall_id),
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> list[MovieSession]:
+    if session_date is not None:
+        return MovieSession.objects.filter(show_time__date=session_date)
+    return MovieSession.objects.all()
```
ArsenPidhoretskyi, line 19:
```suggestion
    movie_sessions = MovieSession.objects.all()
    if session_date is not None:
        movie_sessions = movie_sessions.filter(show_time__date=session_date)
    return movie_sessions

```

services/movie_session.py
```diff
@@ -0,0 +1,43 @@
...
+        session_id: int,
+        show_time: str = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None,
+) -> None:
+    movie_session = MovieSession.objects.get(id=session_id)
+    if show_time is not None:
+        movie_session.show_time = show_time
+    if movie_id is not None:
+        movie_session.movie = Movie.objects.get(id=movie_id)
```
ArsenPidhoretskyi, line 36: Why you need to get object when you can use movie_id = movied_id
CinemaHall have the same problem

ArsenPidhoretskyi (changes requested):

PR #446, opened 2023-09-01
Develop

VasylTurok:


db/models.py
```diff
@@ -14,3 +14,41 @@ class Actor(models.Model):
...
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(
+        to="Actor",
+        related_name="movies"
+    )
+    genres = models.ManyToManyField(
+        to="Genre",
+        related_name="genres"
```
pavlejviki, line 28: related_name="genres"  is a bit confusing. If you were to access this related name from a Genre object, it would suggest that you would get a QuerySet of Genre objects, which is not the case. What you actually get is a QuerySet of Movie objects that belong to that genre.

A more appropriate related_name for the genres field is "movies"

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #450, opened 2023-09-08
Solution

arinaoltarzhevska:


db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField("Actor")
+    genres = models.ManyToManyField("Genre")
```
Yurnerroo, line 23: add `related_name`

services/movie_session.py
```diff
@@ -0,0 +1,49 @@
...
+    if movie_id:
+        session_to_update.movie_id = movie_id
+    if cinema_hall_id:
+        session_to_update.cinema_hall_id = cinema_hall_id
+    session_to_update.save()
+    return session_to_update
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    MovieSession.objects.get(id=session_id).delete()
```
Yurnerroo, line 49: You can use `get_movie_session_by_id ` here

db/migrations/0002_cinemahall_movie_moviesession.py
```diff
@@ -0,0 +1,42 @@
...
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('db', '0001_initial'),
+    ]
```
Yurnerroo, line 11: change ' to " using `Black` module

services/movie.py
```diff
@@ -0,0 +1,46 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None,
```
Yurnerroo, line 8: change `list[int] = None` to `Optional[list[int]] = None` everywhere. If it accepts `None` as default, it must be `Optional`.

services/movie.py
```diff
@@ -0,0 +1,46 @@
...
+            actors__id__in=actors_ids,
+        )
+    elif genres_ids:
+        queryset = queryset.filter(
+            genres__id__in=genres_ids,
+        )
+    elif actors_ids:
+        queryset = queryset.filter(
+            actors__id__in=actors_ids,
+        )
```
Yurnerroo, line 24: refactor using two if statements: `if genres_ids:` and `if actors_ids:`. Get rid of `if genres_ids and actors_ids:` statemtent.

services/movie_session.py
```diff
@@ -0,0 +1,49 @@
+from django.db.models import QuerySet
+
+from db.models import MovieSession
+import datetime
+
+
+def create_movie_session(
+        movie_show_time: datetime,
```
Yurnerroo, line 8: annotation must be datetime.datetime, not datetime module

Yurnerroo (changes requested):

services/movie_session.py
```diff
@@ -0,0 +1,51 @@
...
+        session_to_update.movie_id = movie_id
+    if cinema_hall_id:
+        session_to_update.cinema_hall_id = cinema_hall_id
+    session_to_update.save()
+    return session_to_update
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    movie_session = get_movie_session_by_id(session_id)
+    if movie_session is not None:
```
vasylhnatiuk, line 50: This line doesn't make sense. The code `MovieSession.objects.get(id=movie_session_id)` in Django typically does not return None if no matching record is found. Instead, it raises a `MovieSession.DoesNotExist` exception if there is no record in the database with the specified id.

vasylhnatiuk (changes requested):

services/movie_session.py
```diff
@@ -0,0 +1,51 @@
...
+) -> MovieSession:
+    return MovieSession.objects.create(
+        show_time=movie_show_time,
+        cinema_hall_id=cinema_hall_id,
+        movie_id=movie_id
+    )
+
+
+def get_movies_sessions(
+        session_date: str = None,
```
vasylhnatiuk, line 20: add Optional everywhere when argument can be None

vasylhnatiuk (changes requested):

vasylhnatiuk (changes requested):
fix tetss

vasylhnatiuk (approved):

PR #453, opened 2023-09-10
Solution

nyanplague:


db/models.py
```diff
@@ -14,3 +14,36 @@ class Actor(models.Model):
...
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
+
+    def __str__(self) -> str:
+        return f"{self.title}"
```
OlegatorLE, line 26: what sense in f-string here?

db/models.py
```diff
@@ -14,3 +14,36 @@ class Actor(models.Model):
...
+    name = models.CharField(max_length=255)
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
+
+    @property
+    def capacity(self) -> int:
+        return self.rows * self.seats_in_row
+
+    def __str__(self) -> str:
+        return f"{self.name}"
```
OlegatorLE, line 39: and here?

db/models.py
nyanplague: @OlegatorLE really, thank you 👍

services/movie.py
```diff
@@ -0,0 +1,40 @@
...
+    genres_ids: list[int] = None,
+    actors_ids: list[int] = None,
+) -> None:
+    new_movie = Movie.objects.create(
+        title=movie_title, description=movie_description
+    )
+
+    if genres_ids:
+        for genre_id in genres_ids:
+            new_movie.genres.add(genre_id)
```
OlegatorLE, line 36: You can use set instead add to simplefy your code

services/movie_session.py
```diff
@@ -0,0 +1,48 @@
...
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id,
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
+    if session_date:
+        year, month, day = session_date.split("-")
+        return MovieSession.objects.filter(
+            show_time__year=year, show_time__month=month, show_time__day=day
```
OlegatorLE, line 20: You can use suffix __date to get access to date of the DateTimeField (read hints in task :) )

OlegatorLE (changes requested):
looks good, but need some fix

services/movie_session.py
nyanplague: @OlegatorLE thank youuu!

nyanplague:
FYI I didn't change tests file, it's black :(

PR #465, opened 2023-09-11
Solution

Kurta97:


services/movie_session.py
```diff
@@ -0,0 +1,56 @@
...
+        show_time: str = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> None:
+    updated_movie_session = MovieSession.objects.get(id=session_id)
+
+    if show_time:
+        updated_movie_session.show_time = show_time
+
+    if movie_id:
```
denys-source, line 46: I think this implicit check will be a problem if `movie_id` or `cinema_hall_id` is 0

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="actors")
```
katryana, line 22: In this case, you should set related_name to "movies"

katryana (approved):

services/movie_session.py
Kurta97: Good advice!

db/models.py
Kurta97: I've done it)

denys-source (approved):

PR #467, opened 2023-09-11
Solution

denys-source:


services/movie.py
```diff
@@ -0,0 +1,39 @@
+from django.db.models import QuerySet
+from db.models import Movie
+
+
+def get_movies(
+    genres_ids: list[int] | None = None, actors_ids: list[int] | None = None
+) -> QuerySet[Movie]:
+    movies: QuerySet = Movie.objects.all()
+
+    if genres_ids is not None:
```
MarianKovalyshyn, line 10: You can use it without 'is not None' and it will work the same

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+
+    if movie_id is not None:
+        sessions.update(movie_id=movie_id)
+
+    if cinema_hall_id is not None:
+        sessions.update(cinema_hall_id=cinema_hall_id)
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    MovieSession.objects.filter(id=session_id).delete()
```
MarianKovalyshyn, line 50: I think it's better to use 'get()' instead of 'filter()' because 'session_id' is unique so we will never get more than 1 object.

MarianKovalyshyn (approved):
Looks very good.

services/movie_session.py
denys-source: Good point. Thank you)

Sofbog (changes requested):

PR #470, opened 2023-09-11
Solution

o-davydova:


settings.py
```diff
@@ -21,6 +21,6 @@
 
 USE_I18N = True
 
-USE_TZ = False
+USE_TZ = True
 
-INSTALLED_APPS = ("db",)
+INSTALLED_APPS = ("db", "django_extensions",)
```
viktoria-rybenchuk, line 26: why did you add this?)

viktoria-rybenchuk (changes requested):
you have failed tests

settings.py
o-davydova: It helped me to check the code step by step during development, now I will clean it up)

services/movie.py
```diff
@@ -0,0 +1,45 @@
...
+        movie_description: str,
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
+) -> None:
+    movie = Movie.objects.create(
+        title=movie_title,
+        description=movie_description
+    )
+
+    if genres_ids is not None:
```
viktoria-rybenchuk, line 41: you can check it like  "if genres". Fix in all places

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
Nice!

PR #472, opened 2023-09-11
Solution py-movie-moviesession-cinemahall

katryana:


ZaichenkoViktoriia (approved):

services/movie.py
```diff
@@ -0,0 +1,35 @@
...
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list = None,
+        actors_ids: list = None
+) -> QuerySet:
+    queryset = Movie.objects.all()
+    if genres_ids is not None:
```
Sofbog, line 11: Can we try to refactor this place and all other places in PR.
`if genres_ids is not None:` do we really need this extra check: `is not None`

Sofbog (changes requested):

services/movie.py
katryana: Explicity is better than implicity?

PR #482, opened 2023-09-21
Solution

AndriiIshchenko:


db/migrations/0001_initial.py
```diff
@@ -1,44 +1,58 @@
...
+    initial = True
+
+    dependencies = [
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='Actor',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
```
vasylhnatiuk, line 18: change ' to "

services/movie_session.py
```diff
@@ -0,0 +1,42 @@
...
+
+    return sessions
+
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    MovieSession.objects.filter(id=session_id).delete()
```
vasylhnatiuk, line 29: use get_movie_session_by_id here

services/movie_session.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def update_movie_session(session_id: int,
+                         show_time: datetime = None,
+                         movie_id: int = None,
+                         cinema_hall_id: int = None) -> None:
+    session = MovieSession.objects.filter(id=session_id)
```
vasylhnatiuk, line 36: the same

vasylhnatiuk (changes requested):

services/movie.py
```diff
@@ -0,0 +1,29 @@
+from django.db.models import QuerySet
+from db.models import Movie
+
+
+def get_movies(genres_ids: list[int] = None,
+               actors_ids: list[int] = None) -> QuerySet:
```
Grayder0152, line 6: add Optional typing

services/movie.py
```diff
@@ -0,0 +1,29 @@
+from django.db.models import QuerySet
+from db.models import Movie
+
+
+def get_movies(genres_ids: list[int] = None,
+               actors_ids: list[int] = None) -> QuerySet:
+    queryset = Movie.objects.all()
+    if genres_ids:
```
Grayder0152, line 8: is not None

Grayder0152 (changes requested):

services/movie_session.py
AndriiIshchenko: why? i need "query" not "object" to use .update() i wanted to avoid working with object (than we need use save()) its complicates code. On my opinion.

vasylhnatiuk (approved):

PR #483, opened 2023-09-22
Solution

Daniil-Pankieiev:


vshvanska (approved):

services/movie.py
```diff
@@ -0,0 +1,36 @@
...
+    queryset = Movie.objects.all()
+    if genres_ids:
+        queryset = queryset.filter(genres__id__in=genres_ids)
+    if actors_ids:
+        queryset = queryset.filter(actors__id__in=actors_ids)
+    return queryset
+
+
+def get_movie_by_id(movie_id: int) -> QuerySet:
+    return Movie.objects.get(id=movie_id)
```
LiudmylaKulinchenko, line 19: Remember:`.all()`, `.filter()` returns QuerySet`.get()`, `.create()` returns model class, `-> User` for. example

LiudmylaKulinchenko (changes requested):

services/movie.py
Daniil-Pankieiev: Fixed

LiudmylaKulinchenko (approved):
Well done 👏

PR #487, opened 2023-09-26
Develop

3lobites:


services/movie_session.py
```diff
@@ -0,0 +1,53 @@
...
+    work_cinema_hall = CinemaHall.objects.get(id=cinema_hall_id)
+    new_movie_session = MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie=work_movie,
+        cinema_hall=work_cinema_hall
+    )
+    return new_movie_session
+
+
+def get_movies_sessions(session_date: str = None) -> List[MovieSession]:
```
Bohdan-Salamakha, line 22: use typehint like: Optional[YOUR_VALUE_TYPE] = None

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
...
+
+def get_movie_session_by_id(movie_session_id: MovieSession) -> Movie:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: datetime = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
```
Bohdan-Salamakha, line 40: use typehint like: Optional[YOUR_VALUE_TYPE] = None

Bohdan-Salamakha (changes requested):

YukihiroSM (approved):
Looks good! Thanks for fixing typehints!

PR #490, opened 2023-10-01
'Solution'

ArtemPutsenko:


services/movie_session.py
```diff
@@ -0,0 +1,61 @@
...
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return get_object_or_404(MovieSession, pk=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: datetime = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
```
Bohdan-Salamakha, line 41: use typehint like: Optional[YOUR_VALUE_TYPE] = None

services/movie_session.py
```diff
@@ -0,0 +1,61 @@
...
+        if show_time:
+            movie_session.show_time = show_time
+        if movie_id is not None:
+            movie_session.movie_id = movie_id
+        if cinema_hall_id is not None:
+            movie_session.cinema_hall_id = cinema_hall_id
+
+        movie_session.save()
+    except MovieSession.DoesNotExist:
+        pass
```
Bohdan-Salamakha, line 55: provide some logs not just 'pass'

services/movie_session.py
```diff
@@ -0,0 +1,61 @@
...
+) -> MovieSession:
+    try:
+        movie_session = MovieSession.objects.create(
+            show_time=movie_show_time,
+            movie_id=movie_id,
+            cinema_hall_id=cinema_hall_id
+        )
+        return movie_session
+    except Exception as e:
+        return e
```
Bohdan-Salamakha, line 23: u don't need to return the exception, just print some info

Bohdan-Salamakha (changes requested):

Bohdan-Salamakha (approved):

PR #491, opened 2023-10-03
Develop

Shadowtion:
completed


services/movie.py
```diff
@@ -0,0 +1,41 @@
...
+        return Movie.objects.all()
+    if genres_ids is not None and actors_ids is not None:
+        return Movie.objects.filter(
+            genres__in=genres_ids,
+            actors__in=actors_ids
+        )
+    if genres_ids is not None:
+        return Movie.objects.filter(genres__in=genres_ids)
+    if actors_ids is not None:
+        return Movie.objects.filter(actors__in=actors_ids)
```
YukihiroSM, line 20: You can create a queryset and after that if genres_ids is not None - filter by genres, if actors_ids is not None - filter by actors and the only once return the queryset.  In this case you will have only 2 if statements instead of 4.

YukihiroSM (changes requested):

services/movie.py
Shadowtion: fixed

db/models.py
```diff
@@ -14,3 +15,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField(max_length=255, null=True)
+    actors = models.ManyToManyField(Actor, related_name="movies")
+    genres = models.ManyToManyField(Genre, related_name="movie_genres")
```
ArsenPidhoretskyi, line 24:
```suggestion
    genres = models.ManyToManyField(Genre, related_name="movies")
```

db/models.py
```diff
@@ -14,3 +15,35 @@ class Actor(models.Model):
...
+        return self.rows * self.seats_in_row
+
+    def __str__(self) -> str:
+        return self.name
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField(default=django.utils.timezone.now)
+    cinema_hall = models.ForeignKey(CinemaHall, on_delete=models.CASCADE)
+    movie = models.ForeignKey(Movie, on_delete=models.CASCADE)
```
ArsenPidhoretskyi, line 46: Add related names

services/movie_session.py
```diff
@@ -0,0 +1,45 @@
...
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> None:
+    session = MovieSession.objects.filter(id=session_id)
+    if show_time is not None:
+        session.update(show_time=show_time)
+    if movie_id is not None:
+        session.update(movie_id=movie_id)
+    if cinema_hall_id is not None:
+        session.update(cinema_hall_id=cinema_hall_id)
```
ArsenPidhoretskyi, line 41: You can do 3 queries. Optimize to use 2(get + save) or one

ArsenPidhoretskyi (changes requested):

YukihiroSM (approved):
Looks good to me!

PR #494, opened 2023-10-06
Solution

numerodiciannove:


services/movie.py
```diff
@@ -0,0 +1,53 @@
...
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
+) -> QuerySet:
+    movies = Movie.objects.all()
+
+    if genres_ids and actors_ids:
+        movies = movies.filter(
+            genres__id__in=genres_ids, actors__id__in=actors_ids).distinct()
```
ArsenPidhoretskyi, line 14: This if is redundant. No need to use elif it can be separated if statements

services/movie.py
```diff
@@ -0,0 +1,53 @@
...
+        if genres_ids:
+            new_movie.genres.set(genres_ids)
+
+        if actors_ids:
+            new_movie.actors.set(actors_ids)
+
+    except ValidationError as e:
+        raise ValueError("Validation Error: " + str(e))
+    except Exception as e:
+        raise ValueError("An error occurred: " + str(e))
```
ArsenPidhoretskyi, line 53: Why you need this?

services/movie.py
```diff
@@ -0,0 +1,53 @@
...
+        movies = movies.filter(actors__id__in=actors_ids).distinct()
+
+    return movies
+
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    try:
+        return Movie.objects.get(id=movie_id)
+    except ObjectDoesNotExist:
+        raise ValueError("Movie with the provided ID does not exist.")
```
ArsenPidhoretskyi, line 29: Use get_object_or_404

services/movie_session.py
```diff
@@ -0,0 +1,48 @@
...
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> None:
+    session = MovieSession.objects.filter(id=session_id)
+    if show_time is not None:
+        session.update(show_time=show_time)
+    if movie_id is not None:
+        session.update(movie_id=movie_id)
+    if cinema_hall_id is not None:
+        session.update(cinema_hall_id=cinema_hall_id)
```
ArsenPidhoretskyi, line 42: Instead of 3 queries, you can use .get, set all values, .save()

ArsenPidhoretskyi (changes requested):
Tests failed. What about migrations?

db/migrations/0002_cinemahall_movie_moviesession.py
```diff
@@ -0,0 +1,42 @@
...
+
+    operations = [
+        migrations.CreateModel(
+            name='CinemaHall',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('name', models.CharField(max_length=255)),
+                ('rows', models.IntegerField()),
+                ('seats_in_row', models.IntegerField()),
+            ],
```
Yurnerroo, line 21: change ' to " with Black

Yurnerroo (changes requested):

Yurnerroo (approved):

PR #502, opened 2023-10-15
Solution

MaksymUK:


services/movie.py
```diff
@@ -0,0 +1,40 @@
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
EdAlekseiev, line 6: If argument can be None you must show it in type hint

EdAlekseiev, line 6:
```suggestion
        genres_ids: list[int] | None = None,
        actors_ids: list[int] | None = None
```

services/movie.py
```diff
@@ -0,0 +1,40 @@
...
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
EdAlekseiev, line 27: The same comment as above

services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: str = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
```
EdAlekseiev, line 36: The same comment as above

services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+    return MovieSession.objects.create(
+        show_time=movie_show_time,
+        cinema_hall_id=cinema_hall_id,
+        movie_id=movie_id
+    )
+
+
+def get_movies_sessions(
+        session_date: datetime.date = None
+) -> MovieSession:
```
EdAlekseiev, line 19: Incorrect type hint for return value. This function returns QuerySet of MovieSession

services/movie.py
```diff
@@ -0,0 +1,40 @@
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
+) -> Movie:
```
EdAlekseiev, line 7: Incorrect type hint for return value. This function returns QuerySet of Movie.

EdAlekseiev (changes requested):

services/movie_session.py
```diff
@@ -0,0 +1,55 @@
...
+        show_time: str | None = None,
+        movie_id: int | None = None,
+        cinema_hall_id: int | None = None
+) -> None:
+    movie_session_to_update = MovieSession.objects.get(id=session_id)
+    if show_time:
+        movie_session_to_update.show_time = show_time
+
+    if movie_id is not None:
+        movie_session_to_update.movie = Movie.objects.get(id=movie_id)
```
ArsenPidhoretskyi, line 44: You can use: movie_session_to_update.movie_id = movie_id
No need to do additional queries

ArsenPidhoretskyi: Check in all places

services/movie_session.py
```diff
@@ -0,0 +1,55 @@
...
+
+    if cinema_hall_id is not None:
+        movie_session_to_update.cinema_hall = (
+            CinemaHall.objects.get(id=cinema_hall_id))
+
+    movie_session_to_update.save()
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    movie_session_to_delete = MovieSession.objects.get(id=session_id)
```
ArsenPidhoretskyi, line 54: You can reuse get_movie_session_by_id

ArsenPidhoretskyi (changes requested):

services/movie_session.py
```diff
@@ -0,0 +1,57 @@
...
+) -> QuerySet[MovieSession]:
+    return MovieSession.objects.create(
+        show_time=movie_show_time,
+        cinema_hall_id=cinema_hall_id,
+        movie_id=movie_id
+    )
+
+
+def get_movies_sessions(
+        session_date: datetime.date = None
```
YukihiroSM, line 20: If the default value is None - please, use Optional[]

YukihiroSM (changes requested):
One tiny comment, else looks good to me!

services/movie_session.py
MaksymUK: done

services/movie_session.py
EdAlekseiev: Stiil not fixed

EdAlekseiev (changes requested):

NikOneZ1 (approved):
Good job!

PR #513, opened 2023-10-27
Solution

ybondaren:


services/movie.py
```diff
@@ -0,0 +1,45 @@
...
+        movie_queryset = movie_queryset.filter(actors__id__in=actors_ids)
+
+    return movie_queryset
+
+
+def get_movie_by_id(movie_id: int) -> Optional[Movie]:
+    try:
+        return Movie.objects.get(id=movie_id)
+    except Movie.DoesNotExist:
+        return None
```
ArsenPidhoretskyi, line 26: Use get_object_or_404

ArsenPidhoretskyi (changes requested):

services/movie.py
```diff
@@ -0,0 +1,43 @@
...
+
+from django.db.models import QuerySet
+from django.shortcuts import get_object_or_404
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: List[int] = None,
+        actors_ids: List[int] = None
```
vsmutok, line 11: also better to use `Optional` in annotation

vsmutok (approved):
👍

PR #515, opened 2023-10-28
Solution_movie_sessions

nsdNite:


services/movie.py
```diff
@@ -0,0 +1,56 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list = None,
+        actors_ids: list = None,
```
vsmutok, line 8: use Optional in annotation

vsmutok: fix everywhere

vsmutok (changes requested):

vsmutok (approved):
Great  👍

PR #518, opened 2023-10-31
Solution

MatsalakO:


db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
...
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="actors")
+    genres = models.ManyToManyField(Genre, related_name="genres")
+
```
pavlejviki, line 24: related names are wrong because when accessing an actor, actor.actors would give you movies, not actors, and similarly for genres.

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
...
+    def __str__(self) -> str:
+        return f"{self.title}"
```
pavlejviki, line 26: no need to use f-string, simply return self.title

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
...
+    name = models.CharField(max_length=255)
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
+
+    @property
+    def capacity(self) -> int:
+        return self.rows * self.seats_in_row
+
+    def __str__(self) -> str:
+        return f"{self.name}"
```
pavlejviki, line 39: no need to use f-string

services/movie.py
```diff
@@ -0,0 +1,49 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
```
pavlejviki, line 7: Optional[List[int]] is used to type hint that the parameter can be a list of integers or None. fix in all places

services/movie.py
```diff
@@ -0,0 +1,49 @@
...
+        actors_ids: list[id] = None
```
pavlejviki, line 8:
```suggestion
        actors_ids: list[int] = None
```

services/movie.py
```diff
@@ -0,0 +1,49 @@
...
+
+    if genres_ids is not None:
+        return queryset.filter(
+            genres__id__in=genres_ids
+        )
+
+    if actors_ids is not None:
+        return queryset.filter(
+            actors__id__in=actors_ids
+        )
```
pavlejviki, line 27: code can be simplified, no need for several return statements.

queryset can be filtered based on genres_ids and actors_ids independently, and only if they are provided.

services/movie.py
```diff
@@ -0,0 +1,49 @@
...
+
+    if genres_ids is not None and actors_ids is not None:
+        return queryset.filter(
+            genres__id__in=genres_ids
+        ).filter(
+            actors__id__in=actors_ids
+        )
+
+    if genres_ids is not None:
+        return queryset.filter(
```
pavlejviki, line 20:
```suggestion
queryset = queryset.filter(
```

services/movie.py
```diff
@@ -0,0 +1,49 @@
...
+    queryset = Movie.objects.all()
+
+    if genres_ids is not None and actors_ids is not None:
+        return queryset.filter(
+            genres__id__in=genres_ids
+        ).filter(
+            actors__id__in=actors_ids
+        )
+
+    if genres_ids is not None:
```
pavlejviki, line 19:
```suggestion
    if genres_ids:
```

services/movie.py
```diff
@@ -0,0 +1,49 @@
...
+        return queryset.filter(
+            genres__id__in=genres_ids
+        )
+
+    if actors_ids is not None:
+        return queryset.filter(
```
pavlejviki, line 25:
```suggestion
        queryset =  queryset.filter(
```

services/movie.py
```diff
@@ -0,0 +1,49 @@
...
+        actors_ids: list[id] = None
+) -> QuerySet[Movie]:
+    queryset = Movie.objects.all()
+
+    if genres_ids is not None and actors_ids is not None:
+        return queryset.filter(
+            genres__id__in=genres_ids
+        ).filter(
+            actors__id__in=actors_ids
+        )
```
pavlejviki, line 17: redundant code

services/movie.py
```diff
@@ -0,0 +1,49 @@
...
+
+    if actors_ids is not None:
+        return queryset.filter(
+            actors__id__in=actors_ids
+        )
+    return queryset
+
+
+def get_movie_by_id(movie_id: int) -> QuerySet[Movie]:
```
pavlejviki, line 31: The function get_movie_by_id returns a single Movie instance, not a QuerySet

services/movie_session.py
```diff
@@ -0,0 +1,46 @@
...
+    MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
+
+
+def get_movies_sessions(session_date: datetime = None) -> MovieSession:
+    queryset_sessions = MovieSession.objects.all()
+    if session_date is not None:
```
pavlejviki, line 20:
```suggestion
    if session_date:
```

pavlejviki: remove is not None in all places

db/models.py
pavlejviki: you should leave the attribute, just with the correct value . use "movies"

services/movie.py
pavlejviki: is not fixed

services/movie.py
pavlejviki: is not fixed

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #519, opened 2023-11-01
solution

Shum430:


db/services/movie.py
```diff
@@ -0,0 +1,42 @@
...
+    if not genres_ids and not actors_ids:
+        return Movie.objects.all()
+    elif genres_ids and actors_ids:
+        return Movie.objects.filter(
+            genres__id__in=genres_ids, actors__id__in=actors_ids
+        )
+    elif genres_ids:
+        return Movie.objects.filter(genres__id__in=genres_ids)
+    elif actors_ids:
+        return Movie.objects.filter(actors__id__in=actors_ids)
```
viktoria-rybenchuk, line 19: You can simplify the get_movies function and make it more concise by using conditional expressions (also known as ternary expressions) and chaining filter calls. Here's a more streamlined version of the function that combines the conditions into a single return statement

db/services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+    if movie_id is not None:
+        session.movie_id = movie_id
+    if cinema_hall_id is not None:
+        session.cinema_hall_id = cinema_hall_id
+    session.save()
+    return session
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    session = MovieSession.objects.get(id=session_id)
```
viktoria-rybenchuk, line 53: this is redundant. Use previous function you have created for this logic

db/services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+
+def get_movies_sessions(session_date: datetime = None) -> QuerySet:
+    if session_date:
+        return MovieSession.objects.filter(show_time__date=session_date)
+    else:
+        return MovieSession.objects.values_list(
+            "show_time",
+            "cinema_hall__name",
+            "movie__title"
+        )
```
viktoria-rybenchuk, line 28: You can simplify this code and use a single return statement by leveraging a conditional expression

db/services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: datetime = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> MovieSession:
+    session = MovieSession.objects.get(id=session_id)
+    if show_time is not None:
```
viktoria-rybenchuk, line 42: don't check explicitly for `None`

viktoria-rybenchuk (changes requested):
rollback changes in test_main.py

vsmutok (approved):
👍

PR #520, opened 2023-11-01
Solution

vshvanska:


Daniil-Pankieiev (approved):

services/movie_session.py
```diff
@@ -0,0 +1,44 @@
...
+        movie_session.movie_id = movie_id
+    if cinema_hall_id:
+        movie_session.cinema_hall_id = cinema_hall_id
+
+    movie_session.save()
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    movie_session = get_movie_session_by_id(session_id)
+    movie_session.delete()
```
LiudmylaKulinchenko, line 44: You can just do like this `MovieSession.objects.get(id=session_id).delete()`

db/models.py

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Great!🌷

PR #521, opened 2023-11-02
Movie-MovieSession-CinemaHall Solution

Lyutillis:


db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
...
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
+
+    def __str__(self) -> str:
+        return f"{self.title}"
```
Naz-iv, line 26: in checklist we got recommendation to not use f-string in cases like this, you can return self.title since it's  string

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
...
+    name = models.CharField(max_length=255)
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
+
+    @property
+    def capacity(self) -> int:
+        return self.rows * self.seats_in_row
+
+    def __str__(self) -> str:
+        return f"{self.name}"
```
Naz-iv, line 39: in checklist we got recommendation to not use f-string in cases like this, you can return self.name since it's  string

Naz-iv (approved):
looking good, but below comments need to be implemented

services/movie_session.py
```diff
@@ -0,0 +1,40 @@
...
+                         show_time: datetime = None,
+                         movie_id: int = None,
+                         cinema_hall_id: int = None) -> None:
+    session = MovieSession.objects.filter(pk=session_id)
+    if show_time:
+        session.update(show_time=show_time)
+    if movie_id:
+        session.update(movie=Movie.objects.get(pk=movie_id))
+    if cinema_hall_id:
+        session.update(cinema_hall=CinemaHall.objects.get(pk=cinema_hall_id))
```
LiudmylaKulinchenko, line 36: U can just change values of session here

LiudmylaKulinchenko (changes requested):

db/models.py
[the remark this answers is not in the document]
LiudmylaKulinchenko: you missed that

services/movie.py
```diff
@@ -0,0 +1,32 @@
...
+                 genres_ids: list[int] = None,
+                 actors_ids: list[int] = None) -> None:
+    movie = Movie.objects.create(
+        title=movie_title,
+        description=movie_description
+    )
+    if genres_ids:
+        movie.genres.add(*genres_ids)
+    if actors_ids:
+        movie.actors.add(*actors_ids)
```
LiudmylaKulinchenko, line 32: better use .set(genres_ids)

services/movie.py
```diff
@@ -0,0 +1,32 @@
+from db.models import Movie
+
+
+def get_movies(genres_ids: list[int] = None,
+               actors_ids: list[int] = None) -> list[Movie]:
+    if genres_ids and actors_ids:
+        return Movie.objects.filter(genres__id__in=genres_ids,
+                                    actors__id__in=actors_ids)
```
LiudmylaKulinchenko, line 8: You can remove that. Just not return a querysets in the next conditions, just filter them there

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (changes requested):
You have some comments from Oleksandr

LiudmylaKulinchenko (approved):
Good job 👏

PR #525, opened 2023-11-02
Solution

mykolamateichuk:


services/movie.py
```diff
@@ -0,0 +1,35 @@
...
+        queryset = queryset.filter(genres__id__in=genres_ids)
+
+    if actors_ids is not None:
+        queryset = queryset.filter(actors__id__in=actors_ids)
+
+    return queryset
+
+
+def get_movie_by_id(movie_id: int) -> QuerySet:
+    return Movie.objects.get(id=movie_id)
```
prochigor, line 19: Are you sure that you return QuerySet when there get, has got only one argument?

services/movie_session.py
```diff
@@ -0,0 +1,49 @@
...
+
+    if session_date is not None:
+        queryset = queryset.filter(
+            show_time__date=session_date
+        )
+
+    return queryset
+
+
+def get_movie_session_by_id(movie_session_id: int) -> QuerySet:
```
prochigor, line 28: the same

services/movie_session.py
```diff
@@ -0,0 +1,49 @@
...
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(session_id: int,
+                         show_time: str = None,
+                         movie_id: int = None,
+                         cinema_hall_id: int = None) -> None:
+    queryset = MovieSession.objects.filter(id=session_id)
+
+    if show_time is not None:
```
prochigor, line 38: why not just if show_time: ?

prochigor (approved):

db/models.py
```diff
@@ -14,3 +14,36 @@ class Actor(models.Model):
...
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall, on_delete=models.CASCADE)
+    movie = models.ForeignKey(Movie, on_delete=models.CASCADE)
+
+    def __str__(self) -> str:
+        return (f"{self.movie.title} "
+                f"{self.show_time.strftime('%Y-%m-%d %H:%M:%S')}")
```
LiudmylaKulinchenko, line 49: `.strftime('%Y-%m-%d %H:%M:%S')}` can be removed, since this is the default string representation of a datetime object (one without fractional seconds).

LiudmylaKulinchenko (changes requested):
Fix according to Ihors comments. Don't forget to fix throughout the code

LiudmylaKulinchenko (approved):
Great job! ☀

PR #526, opened 2023-11-03
Solution

prochigor:


D1st3f (approved):
Looks clean!

LiudmylaKulinchenko (changes requested):

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="actors")
+    genres = models.ManyToManyField(Genre, related_name="genres")
```
So, change it for movies

prochigor: Is it ok, when both of them are named the same?

fedorov-s-od (approved):

db/models.py
LiudmylaKulinchenko: Of course! Check an example of a backward relations another time, please
`class Movie(models.Model):
    actors = models.ManyToManyField(Actor, related_name="movies")

    def __str__(self) -> str:
        return self.title`

So here I will use it like this `actor_object.movies.all()` and as a result have all movies that related to actor. Thats why it is correct

LiudmylaKulinchenko: So here I will use it like this `actor_object.movies.all()` and as a result have all movies that related to actor. Thats why it is correct

LiudmylaKulinchenko: Sorry, there is smth wrong with comment, but you can see your Movies model in your code)

PR #528, opened 2023-11-03
Solution

RabenkoYevhenii:


services/movie.py
```diff
@@ -0,0 +1,38 @@
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
+) -> Movie:
```
SevKrok, line 7: Here returns `QuerySet`

SevKrok (approved):

services/movie.py
SevKrok: But all else is good ;)

db/models.py

services/cinema_hall.py
```diff
@@ -0,0 +1,17 @@
...
+def get_cinema_halls() -> CinemaHall:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(
+        hall_name: str,
+        hall_rows: int,
+        hall_seats_in_row: int
+) -> CinemaHall:
+    return CinemaHall.objects.create(
```
LiudmylaKulinchenko, line 13: You don’t need to `return` anything in create/update/delete methods

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+
+def update_movie_session(
+        session_id: int,
+        show_time: str = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> None:
+    movie_session = MovieSession.objects.get(id=session_id)
+
+    if show_time is not None:
```
LiudmylaKulinchenko, line 37: You can just use if show_time
Also in the next conditions

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job! ☀

PR #532, opened 2023-11-03
Solution

Katherine-Greg:


Shatrovskyi (approved):

db/models.py

services/movie.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def get_movie_by_id(movie_id: int) -> QuerySet[Movie]:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(movie_title: str,
+                 movie_description: str,
+                 genres_ids: list[int] = None,
+                 actors_ids: list[int] = None) -> QuerySet[Movie]:
```
LiudmylaKulinchenko, line 28: You don’t need to `return` anything in create/update/delete methods

fsocie7y (approved):

PR #533, opened 2023-11-04
Solution

Shatrovskyi:


kostomeister (approved):
good jobs

db/models.py
```diff
@@ -14,3 +14,43 @@ class Actor(models.Model):
...
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall,
+                                    on_delete=models.SET_NULL,
+                                    null=True,
+                                    related_name="movie_session")
+    movie = models.ForeignKey(Movie,
+                              on_delete=models.SET_NULL,
+                              null=True,
+                              related_name="movie_session")
```
LiudmylaKulinchenko, line 53: do not agree, why do you use SET_NULL here? You should remove movie_sessions if there are no hall or movie anymore

services/cinema_hall.py
```diff
@@ -0,0 +1,19 @@
...
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> QuerySet:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(hall_name: str,
+                       hall_rows: int,
+                       hall_seats_in_row: int) -> CinemaHall:
```

services/movie.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def get_movies(genres_ids: List[int] = None,
+               actors_ids: List[int] = None) -> QuerySet:
+
+    movies_queryset = Movie.objects.all()
+
+    if genres_ids and actors_ids:
+        movies_queryset = movies_queryset.filter(genres__id__in=genres_ids,
+                                                 actors__id__in=actors_ids)
```
LiudmylaKulinchenko, line 13: You can skip this condition, just use if instead of elif in the next two

LiudmylaKulinchenko (changes requested):

Shatrovskyi:
Now, It looks much better.

LiudmylaKulinchenko (approved):
Good job! ☀

PR #536, opened 2023-11-05
Solution

Andrii-Silenko:


services/cinema_hall.py
```diff
@@ -0,0 +1,18 @@
...
+def create_cinema_hall(hall_name: str,
+                       hall_rows: int,
+                       hall_seats_in_row: int) -> CinemaHall:
+    new_cinema_hall = CinemaHall.objects.create(
+        name=hall_name,
+        rows=hall_rows,
+        seats_in_row=hall_seats_in_row
+    )
+
+    return new_cinema_hall
```
SelFisher1488, line 18: as i know, there no need to return model when you create it without optional info

SelFisher1488 (approved):
good code, motivates to continue coding!

db/models.py
```diff
@@ -14,3 +14,39 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="actors")
+    genres = models.ManyToManyField(Genre, related_name="genres")
```
So, change it for movies

db/models.py
```diff
@@ -14,3 +14,39 @@ class Actor(models.Model):
...
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall,
+                                    on_delete=models.CASCADE,
+                                    related_name="cinema_halls")
+    movie = models.ForeignKey(Movie,
+                              on_delete=models.CASCADE,
+                              related_name="movies")
```
LiudmylaKulinchenko, line 49: Don't forget to change also here

services/cinema_hall.py
```diff
@@ -0,0 +1,18 @@
...
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> QuerySet:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(hall_name: str,
+                       hall_rows: int,
+                       hall_seats_in_row: int) -> CinemaHall:
```

services/movie.py
```diff
@@ -0,0 +1,37 @@
+from db.models import Movie
+from django.db.models.query import QuerySet
+
+
+def get_movies(genres_ids: list[int] = None,
+               actors_ids: list[int] = None) -> QuerySet:
+    queryset = Movie.objects.all()
+
+    if genres_ids is not None:
```
LiudmylaKulinchenko, line 9: No need to add `is not None`. What if genres_ids wiil be an empty list? Fix everywhere

LiudmylaKulinchenko (changes requested):

PR #538, opened 2023-11-05
Solution

Wonsky1:


oleksandr-pupchenko (approved):

services/movie.py
```diff
@@ -0,0 +1,42 @@
...
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: List[int] = None,
+        actors_ids: List[int] = None
+) -> django.db.models.QuerySet:
+    queryset = Movie.objects.all()
+
+    if genres_ids is not None:
```
LiudmylaKulinchenko, line 12: you can just check if genres_ids
Same in the next condition

LiudmylaKulinchenko: fix everywhere

services/movie.py
```diff
@@ -0,0 +1,42 @@
...
+
+    if actors_ids is not None:
+        queryset = queryset.filter(actors__id__in=actors_ids)
+
+    return queryset
+
+
+def get_movie_by_id(
+        movie_id: int
+) -> django.db.models.QuerySet:
```
LiudmylaKulinchenko, line 23: `from django.db.models import QuerySet` Don't import whole module

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job! ☀

PR #539, opened 2023-11-05
Solution

DenPrislipskyi:


sakhaline (approved):

db/models.py

services/cinema_hall.py
```diff
@@ -0,0 +1,16 @@
...
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> QuerySet:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(hall_name: str,
+                       hall_rows: int,
+                       hall_seats_in_row: int) -> CinemaHall:
```

LiudmylaKulinchenko (changes requested):

services/movie.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(movie_title: str,
+                 movie_description: str,
+                 genres_ids: list[int] = None,
+                 actors_ids: list[int] = None) -> Movie:
```
LiudmylaKulinchenko, line 23: You don’t need to `return` anything in create/update/delete methods. Don't forget to fix everywhere

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job! ☀

PR #542, opened 2023-11-05
solution

oleksandr-pupchenko:


db/models.py
```diff
@@ -14,3 +14,41 @@ class Actor(models.Model):
...
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True)
+
+    def __str__(self) -> str:
+        return self.title
+
+    actors = models.ManyToManyField(Actor, related_name="actors")
+    genres = models.ManyToManyField(Genre, related_name="genres")
```
vshvanska, line 27: the related_name parameter is used to specify the name of a link from one model to another. Now you are trying to get movies from actors named "actors". "Movies" will be more convenient.

services/cinema_hall.py
```diff
@@ -0,0 +1,40 @@
...
+    cinema_hall = CinemaHall.objects.create(
+        name=hall_name,
+        rows=hall_rows,
+        seats_in_row=hall_seats_in_row
+    )
+
+    return cinema_hall
+
+
+def create_movie(
```
vshvanska, line 23: There is no need to repeat this function. It should be only in movie.py

vshvanska (changes requested):

fsocie7y (changes requested):
all checks should be passed
please fix ur imports

fsocie7y (approved):
GJ 👍

PR #545, opened 2023-11-06
Solution

katyakoziar:


db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="actors")
+    genres = models.ManyToManyField(Genre, related_name="genres")
```
So, change it for movies

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
...
+
+    @property
+    def capacity(self) -> int:
+        return self.rows * self.seats_in_row
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall, on_delete=models.PROTECT)
+    movie = models.ForeignKey(Movie, on_delete=models.PROTECT)
```
LiudmylaKulinchenko, line 45: do not agree, why do you use PROTECT here?

services/movie.py
```diff
@@ -0,0 +1,41 @@
...
+    movies = Movie.objects.all()
+
+    if genres_ids and actors_ids:
+        movies = movies.filter(
+            Q(genres__in=genres_ids) & Q(actors__in=actors_ids)
+        )
+    elif genres_ids:
+        movies = movies.filter(genres__in=genres_ids)
+    elif actors_ids:
+        movies = movies.filter(actors__in=actors_ids)
```
LiudmylaKulinchenko, line 16: You can remove first check, just use `if` instead of elif

services/movie.py
```diff
@@ -0,0 +1,41 @@
...
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(pk=movie_id)
+
+
+def create_movie(
+    movie_title: str,
+    movie_description: str,
+    genres_ids: list = None,
+    actors_ids: list = None
+) -> Movie:
```
LiudmylaKulinchenko, line 30: You don’t need to `return` anything in create/update/delete methods

LiudmylaKulinchenko (changes requested):

db/models.py
katyakoziar: Perhaps I was a little mistaken. Is it appropriate to use CASCADE here?

PR #546, opened 2023-11-06
solution

roffi37:


OlehShumov (approved):

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="actors")
+    genres = models.ManyToManyField(Genre, related_name="genres")
```
So, change it for movies

services/cinema_hall.py
```diff
@@ -0,0 +1,17 @@
...
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> QuerySet:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(hall_name: str,
+                       hall_rows: int,
+                       hall_seats_in_row: int) -> QuerySet:
```

services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+        movie.show_time = show_time
+        movie.save()
+
+    if movie_id:
+        movie.movie_id = movie_id
+        movie.save()
+
+    if cinema_hall_id:
+        movie.cinema_hall_id = cinema_hall_id
+        movie.save()
```
LiudmylaKulinchenko, line 47: You can save once, outside of conditions, just before return

services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+
+    return movie
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    movie = MovieSession.objects.get(id=session_id)
+    movie.delete()
```
LiudmylaKulinchenko, line 54: You can rewrite in one line like this `MovieSession.objects.get(id=session_id).delete()`

LiudmylaKulinchenko (changes requested):

services/movie_session.py
roffi37: Can i do it in return?

LiudmylaKulinchenko (approved):
Great job 🌺

PR #547, opened 2023-11-06
Solution/Alex

AlexIzmashkin:


services/movie.py
```diff
@@ -0,0 +1,38 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
EdAlekseiev, line 8: If argument can be `None` you must show this fact in type hint

EdAlekseiev, line 8:
```suggestion
        genres_ids: list[int] | None = None,
        actors_ids: list[int] | None= None

```

services/movie.py
```diff
@@ -0,0 +1,38 @@
...
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
EdAlekseiev, line 27: The same object as above

services/movie_session.py
```diff
@@ -0,0 +1,60 @@
...
+
+    return MovieSession.objects.create(
+        show_time=show_time,
+        cinema_hall=cinema_hall,
+        movie=movie
+    )
+
+
+def get_movies_sessions(
+        session_date: datetime = None
```
EdAlekseiev, line 29: The same object as above

services/movie_session.py
```diff
@@ -0,0 +1,60 @@
...
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: datetime = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
```
EdAlekseiev, line 46: The same object as above

EdAlekseiev (changes requested):

services/movie_session.py
AlexIzmashkin: corrected

services/movie_session.py
AlexIzmashkin: corrected

services/movie.py
AlexIzmashkin: corrected

services/movie.py
AlexIzmashkin: corrected

YukihiroSM (approved):
LGTM!

PR #552, opened 2023-11-07
Solution

OlehShumov:


skoval7 (approved):

db/models.py

LiudmylaKulinchenko: don't forget to use checklist for that

services/movie.py
```diff
@@ -0,0 +1,38 @@
...
+
+    if genres_ids and actors_ids:
+        queryset = queryset.filter(
+            genres__id__in=genres_ids,
+            actors__id__in=actors_ids
+        )
+    elif genres_ids:
+        queryset = queryset.filter(genres__id__in=genres_ids)
+    elif actors_ids:
+        queryset = queryset.filter(actors__id__in=actors_ids)
```
LiudmylaKulinchenko, line 17: you can skip first condition, just change both `elif` to `if`

services/movie_session.py
```diff
@@ -0,0 +1,40 @@
...
+    return MovieSession.objects.create(show_time=movie_show_time,
+                                       cinema_hall_id=cinema_hall_id,
+                                       movie_id=movie_id)
+
+
+def get_movies_sessions(session_date: datetime.date = None) -> QuerySet:
+    if session_date:
+        return MovieSession.objects.filter(show_time__date=session_date)
+    else:
+        return MovieSession.objects.all()
```
LiudmylaKulinchenko, line 18: no need to use `else` if you have a return in the `if` statement

LiudmylaKulinchenko (changes requested):

PR #555, opened 2023-11-07
'Solution'

fedorov-s-od:


services/movie.py
```diff
@@ -0,0 +1,46 @@
...
+        return Movie.objects.all()
+    elif genres_ids and actors_ids:
+        return Movie.objects.filter(
+            Q(genres__in=genres_ids) & Q(actors__in=actors_ids)
+        )
+    elif genres_ids and not actors_ids:
+        return Movie.objects.filter(genres__in=genres_ids)
+    elif not genres_ids and actors_ids:
+        return Movie.objects.filter(actors__in=actors_ids)
+
```
RVChornyy, line 24: You can use only one return and couple of "if" statements here. You can create one queryset for ocasion when there are no optional arguments. and then filter it if the argument is present. and finally return this queryset.

RVChornyy (changes requested):

services/movie.py
fedorov-s-od: Yep, nice idea! (y) Thanks )

services/movie.py
```diff
@@ -0,0 +1,40 @@
...
+        movie_description: str,
+        genres_ids: Optional[list] = None,
+        actors_ids: Optional[list] = None
+) -> None:
+    movie = Movie.objects.create(
+        title=movie_title, description=movie_description
+    )
+
+    if genres_ids:
+        movie.genres.add(*Genre.objects.filter(id__in=genres_ids))
```
LiudmylaKulinchenko, line 35: better use .set(genres_ids)

LiudmylaKulinchenko (approved):
Well done 🌺

PR #560, opened 2023-11-16
Solution

YegorT:


services/movie.py
```diff
@@ -0,0 +1,42 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
EdAlekseiev, line 8: If an argument can be `None` you must show it in type hint
```suggestion
        genres_ids: list[int] | None = None,
        actors_ids: list[int] | None = None
```

services/movie.py
```diff
@@ -0,0 +1,42 @@
...
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
EdAlekseiev, line 29: The same comment as above

services/movie_session.py
```diff
@@ -0,0 +1,57 @@
...
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: datetime = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
```
EdAlekseiev, line 41: The same comment as above

services/movie_session.py
```diff
@@ -0,0 +1,57 @@
...
+        cinema_hall_id: int
+) -> None:
+    MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
+
+
+def get_movies_sessions(session_date: datetime = None) -> QuerySet:
```
EdAlekseiev, line 22: The same comment as above

EdAlekseiev (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #561, opened 2023-11-17
Solution

SerhiiBudko:


services/movie_session.py
```diff
@@ -0,0 +1,61 @@
...
+def update_movie_session(
+        session_id: int,
+        show_time: Optional[int] = None,
+        movie_id: Optional[int] = None,
+        cinema_hall_id: Optional[int] = None,
+) -> None:
+
+    session = MovieSession.objects.get(id=session_id)
+
+    if show_time is not None:
```
pavlejviki, line 45:
```suggestion
    if show_time:
```

pavlejviki: You can omit the ` is not None` check. fix in all places

pavlejviki (changes requested):

services/movie_session.py
SerhiiBudko: Your's recommendations have been done. Thank's!

vsmutok (approved):
Looks Good 👍

PR #566, opened 2023-11-25
Solution

olgierrd:


db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
...
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="movies")
+    genres = models.ManyToManyField(Genre, related_name="movies")
+
+    def __str__(self) -> str:
+        return f"{self.title}"
```
viktoria-rybenchuk, line 26: why do you need `f-string` here?

services/movie.py
```diff
@@ -0,0 +1,33 @@
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
+) -> Movie:
```
viktoria-rybenchuk, line 7: Wrong type annotations, It should return `Queryset`

services/movie_session.py
```diff
@@ -0,0 +1,59 @@
...
+    return MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
+
+
+def get_movies_sessions(
+        session_date: str = None
+) -> MovieSession:
```
viktoria-rybenchuk, line 18: the same here

services/movie_session.py
```diff
@@ -0,0 +1,59 @@
...
+    if movie_id:
+        session.update(movie=movie_id)
+    if cinema_hall_id:
+        session.update(cinema_hall=cinema_hall_id)
+
+
+def delete_movie_session_by_id(
+        session_id: int
+) -> None:
+    MovieSession.objects.filter(id=session_id).delete()
```
viktoria-rybenchuk, line 59: Don't use `.filter()` when you need to get one item, use `.get()` instead. You  can use previous created method here

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
...
+        return self.rows * self.seats_in_row
+
+    def __str__(self) -> str:
+        return self.name
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall, on_delete=models.CASCADE)
+    movie = models.ForeignKey(Movie, on_delete=models.CASCADE)
```
viktoria-rybenchuk, line 45: add related name

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
GJ🚀

PR #572, opened 2023-11-30
Develop

AnyoneClown:


services/movie.py
```diff
@@ -0,0 +1,50 @@
...
+        queryset = queryset.filter(
+            genres__id__in=genres_ids
+        )
+
+    if actors_ids and not genres_ids:
+        queryset = queryset.filter(
+            actors__id__in=actors_ids
+        )
+
+    return queryset
```
pavlejviki, line 28: You can simplify the get_movies function by first applying one set of criteria (either genres or actors) and then applying the other set of criteria if available. 

First filter the queryset based on genres_ids if it's provided.
Then,  filter the already filtered queryset based on actors_ids if it's provided.

services/movie_session.py
```diff
@@ -0,0 +1,51 @@
...
+
+    if cinema_hall_id:
+        movie_session.update(cinema_hall_id=cinema_hall_id)
+
+    if movie_id:
+        movie_session.update(movie_id=movie_id)
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    MovieSession.objects.filter(id=session_id).delete()
```
pavlejviki, line 51: you can use get_movie_session_by_id method

pavlejviki (changes requested):

PR #573, opened 2023-12-01
Done movie moviesession

OlenaLis123:


services/movie.py
```diff
@@ -0,0 +1,33 @@
+from db.models import Genre, Actor, Movie
+
+
+def get_movies(genres_ids: list[Genre] = None,
+               actors_ids: list[Actor] = None) -> Movie:
```
EdAlekseiev, line 5: If the argument can be `None` you must show it in typehint. Fix it everywhere.
```suggestion
def get_movies(genres_ids: list[Genre] | None = None,
               actors_ids: list[Actor] | None = None) -> Movie:
```

EdAlekseiev, line 5: Invalid type hint for return value. It returns `QuerySet`.

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #574, opened 2023-12-02
Solution

Depaton:


services/movie.py
```diff
@@ -0,0 +1,40 @@
...
+    if genres_ids and actors_ids:
+        queryset = queryset.filter(
+            genres__id__in=genres_ids,
+            actors__id__in=actors_ids
+        )
+    if genres_ids and not actors_ids:
+        queryset = queryset.filter(genres__id__in=genres_ids)
+    if not genres_ids and actors_ids:
+        queryset = queryset.filter(actors__id__in=actors_ids)
+    return queryset
```
pavlejviki, line 20: You can simplify the get_movies function by first applying one set of criteria (either genres or actors) and then applying the other set of criteria if available.

First filter the queryset based on genres_ids if it's provided.
Then, filter the already filtered queryset based on actors_ids if it's provided.

services/movie_session.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
+    queryset = MovieSession.objects.all()
+    if session_date:
+        queryset = queryset.filter(show_time__date=session_date)
+    return queryset
+
+
+def get_movie_session_by_id(movie_session_id: int) -> QuerySet:
```
pavlejviki, line 25: This method does not return a queryset

services/movie_session.py
```diff
@@ -0,0 +1,45 @@
...
+    if show_time:
+        movie_session.update(show_time=show_time)
+    if cinema_hall_id:
+        movie_session.update(cinema_hall_id=cinema_hall_id)
+    if movie_id:
+        movie_session.update(movie_id=movie_id)
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    MovieSession.objects.filter(id=session_id).delete()
```
pavlejviki, line 45: use get_movie_session_by_id

pavlejviki (changes requested):

services/movie_session.py
```diff
@@ -0,0 +1,45 @@
...
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> None:
+    movie_session = MovieSession.objects.filter(id=session_id)
+    if show_time:
+        movie_session.update(show_time=show_time)
+    if cinema_hall_id:
+        movie_session.update(cinema_hall_id=cinema_hall_id)
+    if movie_id:
+        movie_session.update(movie_id=movie_id)
```
ArsenPidhoretskyi, line 41: In this case, you do 3 queries, simplify your code: create a dictionary with parameters to update and call .update only once

ArsenPidhoretskyi (changes requested):

services/cinema_hall.py
```diff
@@ -0,0 +1,18 @@
+from django.db.models import QuerySet
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> QuerySet:
```
Dimosphen1, line 5: You could indicate that you return a `QuerySet` of `CinemaHall` instances in the return type annotation here and similar in other methods as well.

Dimosphen1 (approved):
LGTM, well done!

PR #584, opened 2023-12-13
"Solution"

TROTPOL:


db/models.py
```diff
@@ -14,3 +14,43 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="movie_actors")
+    genres = models.ManyToManyField(Genre, related_name="movie_genres")
```
Dimosphen1, line 23: The `related_name` is not correct for this field, it should not be a case to get all movies from `Actor` by referring to `movie_actors`, and the same for `Genre`. It would be best if you got all movies from `Actor` or `Genre` by using `movies` as a `related_name`, make this change here.

db/models.py
```diff
@@ -14,3 +14,43 @@ class Actor(models.Model):
...
+    cinema_hall = models.ForeignKey(
+        CinemaHall,
+        on_delete=models.PROTECT,
+        related_name="cinema_halls"
+    )
+    movie = models.ForeignKey(
+        Movie,
+        on_delete=models.PROTECT,
+        related_name="movies"
+    )
```
Dimosphen1, line 53: Related names are not selected correctly, change them according to the recommendations from the previous comment.

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+        cinema_hall_id: int
+) -> None:
+    MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
+
+
+def get_movies_sessions(session_date: str | None = None) -> QuerySet:
```
Dimosphen1, line 20: You could indicate that you return a `QuerySet` of `MovieSession` objects in the return type annotation. The same relates to other functions where `QuerySet` is returned.

Dimosphen1 (changes requested):
Several changes were requested.

Dimosphen1 (approved):
LGTM, well done!

PR #590, opened 2023-12-19
py-movie(1)

ostboiko:


services/movie.py
```diff
@@ -0,0 +1,45 @@
+from typing import Optional
+
+from db.models import Movie, Genre, Actor
+
+
+def get_movies(genres_ids: str = None, actors_ids: str = None) -> None:
```
EdAlekseiev, line 6: If an argument can be `None`  you must show it in a type hint. Fix everywhere
```suggestion
def get_movies(genres_ids: str | None = None, actors_ids: str | None = None) -> None:
```

services/movie.py
```diff
@@ -0,0 +1,45 @@
...
+            actors__id__in=actors_ids
+        ).distinct()
+    elif genres_ids:
+        movie = movie.filter(
+            genres__id__in=genres_ids
+        ).distinct()
+    elif actors_ids:
+        movie = movie.filter(
+            actors__id__in=actors_ids
+        )
```
EdAlekseiev, line 20: You can simplify this code
```suggestion
    if genres_ids:
        movie = movie.filter(
            genres__id__in=genres_ids
        ).distinct()
    if actors_ids:
        movie = movie.filter(
            actors__id__in=actors_ids
        )
```

EdAlekseiev (changes requested):

NikOneZ1 (approved):
Good Job!

PR #591, opened 2023-12-20
Solution

aleksandra-prognimak:


db/models.py
```diff
@@ -14,3 +14,45 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="movie")
+    genres = models.ManyToManyField(Genre, related_name="movie")
```
LiudmylaKulinchenko, line 23: the name of the model in plural will be a better option here

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):

PR #605, opened 2024-01-13
Solution

Grotski:


services/movie.py
```diff
@@ -0,0 +1,42 @@
+from django.db.models import QuerySet
+
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
EdAlekseiev, line 9: If argument can be `None` you have to show it in type hint. Fix everywhere.
```suggestion
        genres_ids: list[int] | None = None,
        actors_ids: list[int] | None = None
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #606, opened 2024-01-15
Solution

Anton-Konyk:


services/movie.py
```diff
@@ -0,0 +1,30 @@
+from db.models import Movie
+
+
+def get_movies(genres_ids: list[int] = None,
+               actors_ids: list[int] = None
+               ) -> list[Movie]:
```
EdAlekseiev, line 6: If argument can be None, you must show it in type hint. Fix everywhere, please
```suggestion
def get_movies(genres_ids: list[int] | None = None,
               actors_ids: list[int] | None = None
               ) -> list[Movie]:
```

EdAlekseiev (changes requested):

services/movie.py
Anton-Konyk: def get_movies_sessions(session_date: datetime | None = None
E   TypeError: unsupported operand type(s) for |: 'module' and 'NoneType'

But does not work for datetime?

EdAlekseiev (approved):
Good job

PR #607, opened 2024-01-17
Solution

maxvell4:


services/movie.py
```diff
@@ -0,0 +1,48 @@
...
+        queryset = queryset.filter(
+            genres__id__in=genres_ids,
+            actors__id__in=actors_ids)
+
+    elif genres_ids:
+        queryset = queryset.filter(genres__id__in=genres_ids)
+
+    elif actors_ids:
+        queryset = queryset.filter(actors__id__in=actors_ids)
+
```
pavlejviki, line 23: This check is redundant
 
` if genres_ids and actors_ids:
` 
 Instead of using elif, you can use separate if statements for each condition. 
 
 This approach will apply both filters to the queryset if both genres_ids and actors_ids are present. If only one is present, only the corresponding filter is applied. This way, you don't need the elif statements, and the code is simplified while still achieving the same result.

services/movie.py
```diff
@@ -0,0 +1,48 @@
...
+    return queryset
+
+
+def get_movie_by_id(movie_id: int) -> int:
```
pavlejviki, line 27: does this method return an int?

services/movie.py
```diff
@@ -0,0 +1,48 @@
...
+    return Movie.objects.get(pk=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
+) -> QuerySet[Movie]:
```
pavlejviki, line 36: this method returns a single object

services/movie_session.py
```diff
@@ -0,0 +1,61 @@
+from django.db.models import QuerySet
+
+from db.models import MovieSession
+
+
+def create_movie_session(
+        movie_show_time: str,
+        movie_id: int,
+        cinema_hall_id: int
+) -> QuerySet[MovieSession]:
```
pavlejviki, line 10: this method returns a single object

services/movie_session.py
```diff
@@ -0,0 +1,61 @@
...
+    if session_date:
+        queryset = queryset.filter(show_time__date=session_date)
+    return queryset
+
+
+def get_movie_session_by_id(
+        movie_session_id: int
+) -> QuerySet[MovieSession]:
+
+    queryset = MovieSession.objects.get(pk=movie_session_id)
```
pavlejviki, line 33: creating an additional variable is redundant

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #611, opened 2024-01-23
solution

aiom:


services/movie.py
```diff
@@ -0,0 +1,41 @@
+import init_django_orm  # noqa: F401
+
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(genres_ids: list[int] = None,
+               actors_ids: list[int] = None) -> QuerySet:
```
EdAlekseiev, line 9: If an argument can be `None`, you must show it in type hint
```suggestion
def get_movies(genres_ids: list[int] | None = None,
               actors_ids: list[int] | None = None) -> QuerySet:
```

services/movie.py
```diff
@@ -0,0 +1,41 @@
...
+            actors__id__in=actors_ids
+        ).distinct()
+    if actors_ids:
+        return Movie.objects.filter(
+            actors__id__in=actors_ids
+        ).distinct()
+    if genres_ids:
+        return Movie.objects.filter(
+            genres__id__in=genres_ids
+        ).distinct()
```
EdAlekseiev, line 22: You can simplify this function
```suggestion
    query = Movie.objects.all()
    if actors_ids:
        query = query.filter(actors__id__in=actors_ids)
    if genres_ids:
        query = query.filter(genres__id__in=genres_ids)
    return query.distinct()
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #615, opened 2024-01-26
Solution

nicksetrakov:


services/movie.py
```diff
@@ -0,0 +1,45 @@
...
+        queryset = queryset.filter(actors__id__in=actors_ids)
+
+    return queryset
+
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    try:
+        return Movie.objects.get(id=movie_id)
+    except Movie.DoesNotExist:
+        raise Http404(f"Movie with id {movie_id} does not exist.")
```
LiudmylaKulinchenko, line 25: U can use get_object_or_404 method instead of this construction

LiudmylaKulinchenko (changes requested):

YuriiKindrat (approved):
GJ

PR #625, opened 2024-02-05
Solution

mrve4nick:


services/movie.py
```diff
@@ -0,0 +1,48 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
+) -> QuerySet[Movie] | Movie:
```
Dimosphen1, line 9: Change your type annotation, there is no case when a single `Movie` will be returned, it will be always a `QuerySet` in this function.

services/movie_session.py
```diff
@@ -0,0 +1,48 @@
...
+    if movie_id:
+        movie_session.movie_id = movie_id
+    if cinema_hall_id:
+        movie_session.cinema_hall_id = cinema_hall_id
+
+    movie_session.save()
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    MovieSession.objects.filter(id=session_id).delete()
```
Dimosphen1, line 48: Use the `get_movie_session_by_id` function here.

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
...
+        return self.rows * self.seats_in_row
+
+    def __str__(self) -> str:
+        return f"{self.name}"
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall, on_delete=models.CASCADE)
+    movie = models.ForeignKey(Movie, on_delete=models.CASCADE)
```
Dimosphen1, line 45: Add a correct `related_name` for these fields as well.

Dimosphen1 (changes requested):
Several changes were requested.

Polyakiv-Andrey (approved):
Good job!

PR #630, opened 2024-02-07
Solution

NataliiaKucher:


services/cinema_hall.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+def get_cinema_halls() -> QuerySet | CinemaHall:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(
+        hall_name: str = None,
+        hall_rows: int = None,
+        hall_seats_in_row: int = None
```
EdAlekseiev, line 12: If argument can be `None` you must show it in type hint. Fix everywhere
```suggestion
        hall_name: str | None = None,
        hall_rows: int | None = None,
        hall_seats_in_row: int | None = None
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #632, opened 2024-02-10
Solution

msymonovych:


db/models.py
```diff
@@ -14,3 +14,45 @@ class Actor(models.Model):
...
+        on_delete=models.CASCADE,
+        related_name="movie_sessions",
+        null=True
+    )
+    movie = models.ForeignKey(
+        Movie,
+        on_delete=models.CASCADE,
+        related_name="movie_sessions",
+        null=True
+    )
```
MaxymChyncha, line 55: I think there's no need to use `null=True` here.

services/movie_session.py
```diff
@@ -0,0 +1,57 @@
...
+    if cinema_hall_id:
+        movie_session.cinema_hall_id = cinema_hall_id
+
+    movie_session.save()
+    return movie_session
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    movie = MovieSession.objects.get(id=session_id)
+    movie.delete()
```
MaxymChyncha, line 57: You could make it separately without creating variable `movie`

MaxymChyncha (approved):
Good job!

db/models.py
msymonovych: You're right

PR #644, opened 2024-02-13
Solution

NataliiaKorzh:


services/movie.py
```diff
@@ -0,0 +1,42 @@
...
+
+    if genres_ids:
+        genres = Genre.objects.filter(pk__in=genres_ids)
+        movie.genres.add(*genres)
+
+    if actors_ids:
+        actors = Actor.objects.filter(pk__in=actors_ids)
+        movie.actors.add(*actors)
+
+    return movie
```
VladKyianytsia, line 42: Does this function have to return something?

VladKyianytsia (approved):

PR #645, opened 2024-02-13
Solution1

artemgrishko:


services/movie.py
```diff
@@ -0,0 +1,42 @@
...
+        return Movie.objects.filter(
+            genres__id__in=genres_ids,
+            actors__id__in=actors_ids
+        )
+
+    if genres_ids:
+        return Movie.objects.filter(genres__id__in=genres_ids)
+
+    if actors_ids:
+        return Movie.objects.filter(actors__id__in=actors_ids)
```
aLEKS-e3, line 18: you can get a QuerySet of all movies here, then use 2 `if statements` to filter it, and return the filtered (if filtered) in the end

PR #650, opened 2024-02-13
Solution

MaxOperchuk:


db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
...
+    genres = models.ManyToManyField(Genre)
+
+    def __str__(self) -> str:
+        return self.title
+
+
+class CinemaHall(models.Model):
+    name = models.CharField(max_length=255)
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
```
Dobryk12, line 32: This shit sometimes can be None or empty I quess, use null and blank with True

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
...
+    def __str__(self) -> str:
+        return self.name
+
+    @property
+    def capacity(self) -> int:
+        return self.seats_in_row * self.rows
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
```
Dobryk12, line 43: Same here

services/movie.py
```diff
@@ -0,0 +1,46 @@
...
+    new_film = Movie.objects.create(
+        title=movie_title,
+        description=movie_description,
+    )
+
+    if genres_ids:
+        new_film.genres.set(genres_ids)
+
+    if actors_ids:
+        new_film.actors.set(actors_ids)
```
Dobryk12, line 46: You missed return with new_film

services/movie_session.py
```diff
@@ -0,0 +1,55 @@
...
+
+    if show_time:
+        session_to_update.update(show_time=show_time)
+
+    if movie_id:
+        session_to_update.update(movie=movie_id)
+
+    if cinema_hall_id:
+        session_to_update.update(cinema_hall=cinema_hall_id)
+
```
Dobryk12, line 52: Return session_to_update

Dobryk12 (approved):

services/movie.py
Dobryk12: I mean at the end of full func, sorry I'm noob in Github

PR #655, opened 2024-02-13
Solution

Meksim-fesh:


services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+    if movie_id:
+        session.movie_id = movie_id
+    if cinema_hall_id:
+        session.cinema_hall_id = cinema_hall_id
+    session.save()
+    return session
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    MovieSession.objects.get(id=session_id).delete()
```
vasylhnatiuk, line 50: use get_movie_session_by_id()

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+) -> MovieSession:
+    new_session = MovieSession.objects.create(
+        show_time=movie_show_time,
+        cinema_hall_id=cinema_hall_id,
+        movie_id=movie_id
+    )
+    return new_session
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
```
vasylhnatiuk, line 21: is session_date a str?

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

services/movie_session.py
Meksim-fesh: It should be a str.
- `get_movies_sessions`, takes optional string `session_date` in such
form: "year-month-day" - this is from readme file

PR #657, opened 2024-02-13
solution

romanshupak:


db/models.py
```diff
@@ -14,3 +14,42 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
```
vasylhnatiuk, line 22: add related name

db/models.py
```diff
@@ -14,3 +14,42 @@ class Actor(models.Model):
...
+    genres = models.ManyToManyField(Genre)
```
vasylhnatiuk, line 23: add related name

vasylhnatiuk (changes requested):

services/cinema_hall.py
```diff
@@ -0,0 +1,15 @@
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> CinemaHall:
```
vasylhnatiuk, line 4: it doesn't return an one object

services/movie.py
```diff
@@ -0,0 +1,36 @@
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
+) -> list:
```
vasylhnatiuk, line 7:
```suggestion
) -> Queryset:
```

services/movie_session.py
```diff
@@ -0,0 +1,43 @@
...
+        cinema_hall_id: int
+) -> MovieSession:
+    return MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> MovieSession:
```
vasylhnatiuk, line 15: it doesn't return an one object

vasylhnatiuk: and is sesssion_date is str?

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #662, opened 2024-02-16
Solution

artemlitvenko:


db/models.py
```diff
@@ -14,3 +14,38 @@ class Actor(models.Model):
...
+
+    @property
+    def capacity(self) -> int:
+        return self.rows * self.seats_in_row
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall, on_delete=models.CASCADE)
+    movie = models.ForeignKey(Movie, on_delete=models.CASCADE)
```
MaxymChyncha, line 45: The same situation which I described above.

services/movie_session.py
```diff
@@ -0,0 +1,47 @@
...
+    return MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> None:
+    if session_date:
+        date = timezone.datetime.strptime(session_date, "%Y-%m-%d").date()
```
MaxymChyncha, line 19: You don't need to use `datetime.strptime` here. `session date` already has an appropriate format.

MaxymChyncha (approved):
Overall good but check my commetns)

PR #664, opened 2024-02-18
Solution

achernyshovv:


services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+        cinema_hall_id: Optional[int] = None,
+        show_time: Optional[datetime] = None
+) -> MovieSession:
+    session = MovieSession.objects.get(id=session_id)
+    if movie_id is not None:
+        session.movie_id = movie_id
+    if cinema_hall_id is not None:
+        session.cinema_hall_id = cinema_hall_id
+    if show_time is not None:
+        session.show_time = show_time
```
Anna728560, line 43: You can write just "if name:" instead of "if name is not None:"

Anna728560 (approved):

PR #673, opened 2024-02-26
Solution

vsupruniuk:


services/movie.py
```diff
@@ -0,0 +1,46 @@
...
+        actors_ids: list[int] = None
+) -> QuerySet[Movie]:
+    if not genres_ids and not actors_ids:
+        return Movie.objects.all()
+
+    if actors_ids and not genres_ids:
+        return Movie.objects.filter(actors__id__in=actors_ids)
+
+    if genres_ids and not actors_ids:
+        return Movie.objects.filter(genres__id__in=genres_ids)
```
LiudmylaKulinchenko, line 17: You can make it with only two checks: `if authors_ids` and `if genres_ids`. Do not return in every if block to reach this

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job 🎉

PR #680, opened 2024-03-01
'Solution_Movie_CinemaHall_MovieSession'

vvg04021968:


services/movie_session.py
```diff
@@ -0,0 +1,47 @@
...
+        movie_session.show_time = show_time
+    if movie_id:
+        movie_session.movie_id = movie_id
+    if cinema_hall_id:
+        movie_session.cinema_hall_id = cinema_hall_id
+    movie_session.save()
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    MovieSession.objects.get(id=session_id).delete()
```
vasylhnatiuk, line 47: reuse get_movie_session_by_id here

vasylhnatiuk (changes requested):

services/movie_session.py
vvg04021968: I don't undestand task

vasylhnatiuk (approved):

PR #683, opened 2024-03-03
created new models and services for them

RomanHlodann:


services/movie_session.py
```diff
@@ -0,0 +1,47 @@
...
+    if show_time:
+        movie_session.show_time = show_time
+    if movie_id:
+        movie_session.movie_id = movie_id
+    if cinema_hall_id:
+        movie_session.cinema_hall_id = cinema_hall_id
+    return movie_session.save()
+
+
+def delete_movie_session_by_id(session_id: int) -> dict:
```
Dimosphen1, line 46: This function would return not a `dict` but rather `tuple`. Check what type this function returns and change it accordinly

Dimosphen1 (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #695, opened 2024-03-15
Solution

opminen:


services/movie.py
```diff
@@ -0,0 +1,42 @@
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
```
NikOneZ1, line 5: Also indicate that argument could be also None type everywhere

NikOneZ1 (changes requested):

services/movie_session.py
```diff
@@ -0,0 +1,66 @@
...
+        if movie_session:
+            if show_time:
+                movie_session.show_time = show_time
+            if movie_id:
+                movie_session.movie_id = movie_id
+            if cinema_hall_id:
+                movie_session.cinema_hall_id = cinema_hall_id
+            movie_session.save()
+            return movie_session
+    return None
```
vsmutok, line 56: the condition states:
> update_movie_session, takes session_id, optional show_time, optional movie_id, optional cinema_hall_id. Update movie session with provided session_id and set fields if appropriate values are provided

so there is no need to specify a default value for `session_id`

vsmutok (changes requested):

PR #696, opened 2024-03-15
Implemented py-movie-moviesession-cinemahall

esemesenko:


services/movie.py
```diff
@@ -0,0 +1,35 @@
+from db.models import Movie
+from django.db.models import QuerySet
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
EdAlekseiev, line 7: If argument can be None, you should show it in type anotation
```suggestion
        genres_ids: list[int] | None = None,
        actors_ids: list[int] | None = None
```

services/movie.py
```diff
@@ -0,0 +1,35 @@
...
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
EdAlekseiev, line 25: The same comment as above

services/movie_session.py
```diff
@@ -0,0 +1,51 @@
...
+    movie = Movie.objects.get(id=movie_id)
+    new_movie_session = MovieSession.objects.create(
+        show_time=movie_show_time,
+        cinema_hall=cinema_hall,
+        movie=movie
+    )
+    return new_movie_session
+
+
+def get_movies_sessions(session_date: datetime = None) -> QuerySet:
```
EdAlekseiev, line 21: The same comment as above

services/movie_session.py
```diff
@@ -0,0 +1,51 @@
...
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: datetime = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
```
EdAlekseiev, line 36: The same comment as above

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #706, opened 2024-03-27
Solution

IhorDmytriv:


services/movie_session.py
```diff
@@ -0,0 +1,51 @@
+from django.db.models import QuerySet
+
+from db.models import MovieSession
+
+
+def create_movie_session(
+        movie_show_time: str = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
```
YukihiroSM, line 9: It's better to use Optional[int] typehint if the default value is set to None. Fix it in all functions please.

YukihiroSM (changes requested):
Else looks good to me! Nicely done!

vasylhnatiuk (approved):

PR #715, opened 2024-04-05
Movie, CinemaHall, and MovieSession

Ananiev-Vitalii:


services/movie.py
```diff
@@ -0,0 +1,39 @@
...
+    queryset = Movie.objects.all()
+
+    if genres_ids and actors_ids:
+        queryset = queryset.filter(
+            genres__id__in=genres_ids,
+            actors__id__in=actors_ids)
+    if genres_ids:
+        queryset = queryset.filter(genres__id__in=genres_ids)
+    if actors_ids:
+        queryset = queryset.filter(actors__id__in=actors_ids)
```
EdAlekseiev, line 17:
```suggestion
    if genres_ids:
        queryset = queryset.filter(genres__id__in=genres_ids)
    if actors_ids:
        queryset = queryset.filter(actors__id__in=actors_ids)
```

EdAlekseiev (changes requested):

services/movie.py
Ananiev-Vitalii: @EdAlekseiev Sorry, I don't quite understand where the code needs to be corrected

PR #716, opened 2024-04-05
Solution

RomanNest:


vasylhnatiuk (changes requested):
fix tests

services/movie.py
```diff
@@ -0,0 +1,52 @@
+from django.db.models import QuerySet
+
+from db.models import Genre, Movie, Actor
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None,
```
```suggestion
        genres_ids: list[int] | None = None,
        actors_ids: list[int] | None = None,
```

services/movie.py
```diff
@@ -0,0 +1,52 @@
...
+) -> QuerySet[Movie]:
+    queryset = Movie.objects.all()
+
+    if not genres_ids and not actors_ids:
+        return queryset
```
EdAlekseiev, line 13:
```suggestion
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #722, opened 2024-04-08
'Solution'

donbas777:


db/migrations/0004_rename_descriptions_movie_description.py
Dimosphen1, line 1: It is a good practice to include one migration per app per PR in Django projects since it helps to avoid `migrations` folder overload in the future. Represent all changes done to models in one migration file per app per PR, please

services/cinema_hall.py
```diff
@@ -0,0 +1,20 @@
+from django.db.models import QuerySet
+
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> QuerySet:
```
Dimosphen1, line 6: Indicate a `QuerySet`of what instances you return in the return type annotation

services/movie_session.py
```diff
@@ -0,0 +1,60 @@
...
+    if cinema_hall_id:
+        movie_session.cinema_hall_id = cinema_hall_id
+    movie_session.save()
+
+
+def delete_movie_session_by_id(
+        session_id: int
+) -> None:
+    movie_session = MovieSession.objects.get(id=session_id)
+    movie_session.delete()
```
Dimosphen1, line 60: Use `get_movie_session_by_id` method here and implement solution in one line

Dimosphen1 (changes requested):

EdAlekseiev (approved):

PR #728, opened 2024-04-15
Done

DenZatolokin:


services/movie.py
```diff
@@ -0,0 +1,35 @@
+from django.db.models import QuerySet
+
+from db.models import Genre, Movie, Actor
+
+
+def get_movies(
+        genres_ids: list = None,
+        actors_ids: list = None
```
EdAlekseiev, line 8: If your argument can be `None` you should show it in type hint. Also, specify inner type of list. Fix everywhere, please
```suggestion
        genres_ids: list[int] | None= None,
        actors_ids: list[int] | None = None
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #731, opened 2024-04-16
Solution

TetyanaPavlyuk:


services/movie.py
```diff
@@ -0,0 +1,38 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
```suggestion
        genres_ids: list[int] | None = None,
        actors_ids: list[int] | None = None
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #733, opened 2024-04-17
Solution

ayatsuta:


services/movie_session.py
```diff
@@ -0,0 +1,55 @@
...
+) -> QuerySet[MovieSession]:
+    queryset = MovieSession.objects.all()
+    if session_date:
+        movie_date = datetime.strptime(session_date, "%Y-%m-%d")
+        queryset = queryset.filter(
+            show_time__year=movie_date.year,
+            show_time__month=movie_date.month,
+            show_time__day=movie_date.day
+        )
+    return queryset
```
vsmutok, line 31:
```suggestion
    if session_date:
        return MovieSession.objects.filter(show_time__date=session_date)
    return MovieSession.objects.all()

```

it can be done more easily)

vsmutok (approved):
Good Job!

PR #751, opened 2024-04-23
Solution

SerhiiMar:


fsocie7y (changes requested):
You don’t need to `return` anything in create/update/delete methods

services/movie.py
```diff
@@ -0,0 +1,35 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
```
Oleksl888, line 7: better use Optional here and below wherever None can be expected

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
...
+
+from db.models import MovieSession
+
+
+def create_movie_session(
+        movie_show_time: datetime,
+        movie_id: int,
+        cinema_hall_id: int
+) -> None:
+    return MovieSession.objects.create(
```
Oleksl888, line 13: Adrian already mentioned that you don't need to return newly created instances.

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
...
+        cinema_hall_id=cinema_hall_id
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
+    queryset = MovieSession.objects.all()
+    if session_date:
+        session_date = datetime.strptime(session_date, "%Y-%m-%d")
+        queryset = queryset.filter(
+            show_time__year=session_date.year,
```
Oleksl888, line 25: wouldn't be easier to use `show_time__date` instead?

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
...
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: datetime = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> None:
+    movie_session = MovieSession.objects.get(id=session_id)
```
Oleksl888, line 42: you can rewrite this method for efficiency so you don't have get movie_session and call save on it in case if there are no optional params to update

Oleksl888 (changes requested):

YuriiKindrat (approved):
GJ

PR #752, opened 2024-04-23
'Solution1'

yossosh:


.flake8
Oleksl888, line 1: don't think these changes are necessary

db/models.py
```diff
@@ -14,3 +14,41 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField(null=True, blank=True)
+    actors = models.ManyToManyField(Actor)
```
Oleksl888, line 22: what about related names here?

db/models.py
```diff
@@ -14,3 +14,41 @@ class Actor(models.Model):
...
+    @property
+    def capacity(self) -> int:
+        return self.rows * self.seats_in_row
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall,
+                                    on_delete=models.CASCADE,
+                                    related_name="cinema_hall_id")
```
Oleksl888, line 46: misleading related names here. These names should represent how the movie_session object can be accessed from cinemahall or movie

db/models.py
```diff
@@ -14,3 +14,41 @@ class Actor(models.Model):
...
+    movie = models.ForeignKey(Movie,
+                              on_delete=models.CASCADE,
+                              related_name="movie_id")
+
+    def __str__(self) -> str:
+        movie_title = self.movie.title
+        show_time = self.show_time.strftime("%Y-%m-%d %H:%M:%S")
```
Oleksl888, line 53: no need to save it in a variable if you use it only once

services/cinema_hall.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+def get_cinema_halls() -> QuerySet:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(hall_name: str,
+                       hall_rows: int,
+                       hall_seats_in_row: int) -> CinemaHall:
+    return CinemaHall.objects.create(name=hall_name,
```
Oleksl888, line 13: returning Cinemahall is not a requirement

services/movie.py
```diff
@@ -0,0 +1,33 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(genres_ids: list[int] = None,
```
Oleksl888, line 6: Use Optional here and other similar functions

services/movie.py
```diff
@@ -0,0 +1,33 @@
...
+                 movie_description: str,
+                 genres_ids: list[int] = None,
+                 actors_ids: list[int] = None) -> Movie:
+    movie = Movie.objects.create(title=movie_title,
+                                 description=movie_description)
+    if genres_ids:
+        movie.genres.set(genres_ids)
+    if actors_ids:
+        movie.actors.set(actors_ids)
+    return movie
```
Oleksl888, line 33: no need to return movie here. Double check your other services vs requirements and don't return if unnecessary

services/movie_session.py
```diff
@@ -0,0 +1,45 @@
...
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(session_id: int,
+                         show_time: datetime = None,
+                         movie_id: int = None,
+                         cinema_hall_id: int = None) -> MovieSession:
+    movie_session = MovieSession.objects.get(id=session_id)
```
Oleksl888, line 33: you can rewrite this logic for efficiency, so you don't have to get session and save it in case if you didn't get any optional params. Also returning movie_session is not a requirement

Oleksl888 (changes requested):

.flake8
yossosh: I added ".venv" to exclude because without it flake8 command works for few mins and checks for unnecessary files

db/models.py
yossosh: I did it because the f line was too long before and I didn't know how to split it correctly, so i created a variables

.flake8
Oleksl888, line 1: Please remove I insist :)
You can add these changes locally if it works for you but no need to push them to remote. Here flake runs in github environment and it does all checks correctly

db/models.py
```diff
@@ -14,3 +14,44 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField(null=True, blank=True)
+    actors = models.ManyToManyField(Actor, related_name="movie_actors_id")
```
Oleksl888, line 22: It should be a simple name that represents movies on models of actors or genres. How do you think would be easier to access it:
```
actor.movie_actors_id
OR
actor.movies

File: db/models.py [hunk 9]
```diff
@@ -14,3 +14,44 @@ class Actor(models.Model):
...
+    def capacity(self) -> int:
+        return self.rows * self.seats_in_row
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(
+        CinemaHall,
+        on_delete=models.CASCADE,
+        related_name="movie_session_cinema_hall_id"
```
Oleksl888, line 47: choose more appropriate name here also

services/cinema_hall.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+def get_cinema_halls() -> QuerySet:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(hall_name: str,
+                       hall_rows: int,
+                       hall_seats_in_row: int) -> None:
+    return CinemaHall.objects.create(
```
Oleksl888, line 13: no need to return here

services/movie.py
```diff
@@ -0,0 +1,36 @@
+from django.db.models import QuerySet
+from typing import List, Optional
```
Oleksl888, line 2: check order of imports

services/movie.py
```diff
@@ -0,0 +1,36 @@
...
+        movies_queryset = movies_queryset.filter(genres__id__in=genres_ids)
+
+    if actors_ids:
+        movies_queryset = movies_queryset.filter(actors__id__in=actors_ids)
+
+    return movies_queryset.distinct()
+
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
```
Oleksl888, line 21: it is a convention to use `pk` in ORM. By default primary key is `id` but there can be cases when it's not, so for this we use `pk` alias

services/movie.py
```diff
@@ -0,0 +1,36 @@
...
+                 actors_ids: Optional[List[int]] = None) -> None:
+    movie = Movie.objects.create(
+        title=movie_title,
+        description=movie_description
+    )
+    if genres_ids:
+        movie.genres.set(genres_ids)
+    if actors_ids:
+        movie.actors.set(actors_ids)
+    return movie
```
Oleksl888, line 36: no need to return movie here

services/movie_session.py
```diff
@@ -0,0 +1,48 @@
...
+
+from django.db.models import QuerySet
+
+from db.models import MovieSession
+
+
+def create_movie_session(movie_show_time: datetime,
+                         movie_id: int,
+                         cinema_hall_id: int) -> MovieSession:
+    return MovieSession.objects.create(
```
Oleksl888, line 12: No need to return

Oleksl888 (changes requested):

.flake8
Oleksl888, line 1: I still insist that this file should not have any changes

db/models.py
Oleksl888: Please change related name to more relevant

.flake8
Oleksl888, line 1: remove all diffs from file and push it with next commit

db/models.py
```diff
@@ -14,3 +14,44 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField(null=True, blank=True)
+    actors = models.ManyToManyField(Actor, related_name="actor")
+    genres = models.ManyToManyField(Genre, related_name="genre")
```
Oleksl888, line 23: This must represent how movies will be accessed from genres model

db/models.py
```diff
@@ -14,3 +14,44 @@ class Actor(models.Model):
...
+    def capacity(self) -> int:
+        return self.rows * self.seats_in_row
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(
+        CinemaHall,
+        on_delete=models.CASCADE,
+        related_name="cinema_hall"
```
Oleksl888, line 47: Again - this must represent a link to session object on cinemahall object

services/movie_session.py
```diff
@@ -0,0 +1,48 @@
...
+    movies_sessions_queryset = MovieSession.objects.all()
+    if session_date:
+        movies_sessions_queryset = movies_sessions_queryset.filter(
+            show_time__date=session_date
+        )
+    return movies_sessions_queryset
+
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
```
Oleksl888, line 29: use `pk` here too

Oleksl888 (changes requested):

Oleksl888 (changes requested):

MNDonut (approved):

PR #753, opened 2024-04-23
Developer

semermaks:


db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="movies")
+    genres = models.ManyToManyField(Genre, related_name="genres")
```
Oleksl888, line 23: misleading name here

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
...
+        return self.name
+
+    @property
+    def capacity(self) -> int:
+        return self.rows * self.seats_in_row
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall, on_delete=models.CASCADE)
```
Oleksl888, line 44: missing related names here

services/movie.py
```diff
@@ -0,0 +1,41 @@
+from django.db.models.query import QuerySet
+
+from db.models import Movie, Actor, Genre
+
+
+def get_movies(
+        genres_ids: list[int] = None,
```
Oleksl888, line 7: Use Optional type here and other places with similar params

services/movie.py
```diff
@@ -0,0 +1,41 @@
...
+        actors_ids: list[int] = None
+) -> QuerySet:
+    movies = Movie.objects.all()
+    if genres_ids is not None and actors_ids is not None:
```
Oleksl888, line 11: simplify this to two if statements

services/movie.py
```diff
@@ -0,0 +1,41 @@
...
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: list[int] = None,
+        actors_ids: int = None
```
Oleksl888, line 31: wrong type + need add optional

services/movie.py
```diff
@@ -0,0 +1,41 @@
...
+) -> Movie:
+    movie = Movie.objects.create(
+        title=movie_title,
+        description=movie_description
+    )
+    if genres_ids:
+        movie.genres.set(Genre.objects.filter(id__in=genres_ids))
```
Oleksl888, line 38: Set genres as advised in checklist:
```
player = Player.objects.get(id=1)
skills_ids = [1, 2, 3]

player.skills.set(skills_ids)

File: services/movie.py [hunk 7]
```diff
@@ -0,0 +1,41 @@
...
+    if actors_ids:
+        movie.actors.set(Actor.objects.filter(id__in=actors_ids))
+    return movie
```
Oleksl888, line 41: there is no requirement to return movie

services/movie_session.py
```diff
@@ -0,0 +1,49 @@
...
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: datetime = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> MovieSession:
+    movie_session = MovieSession.objects.get(id=session_id)
```
Oleksl888, line 37: you can rewrite this logic for efficiency, so you don't have to get session and save it in case if you didn't get any optional params. Also returning movie_session is not a requirement

Oleksl888 (changes requested):

services/movie.py
```diff
@@ -0,0 +1,36 @@
+from django.db.models.query import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] | None = None,
```
Oleksl888, line 7: Better use Optional type hint to mark optional params

db/models.py
Oleksl888: hasn't been fixed

db/models.py
Oleksl888: hasn't been fixed

services/movie.py
```diff
@@ -0,0 +1,36 @@
...
+    if genres_ids:
+        movies = movies.filter(genres__in=genres_ids)
+
+    if actors_ids:
+        movies = movies.filter(actors__in=actors_ids)
+    return movies
+
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
```
Oleksl888, line 20: it is a convention to use `pk` in ORM

services/movie.py
```diff
@@ -0,0 +1,36 @@
...
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: list[int] | None = None,
+        actors_ids: list[int] | None = None
+) -> None:
+    movie = Movie.objects.create(
+        title=movie_title,
+        description=movie_description)
+    if genres_ids is not None:
```
Oleksl888, line 32: if genres_ids would be more pythonic

Oleksl888 (changes requested):
Please fix previous comments + more

services/movie_session.py
```diff
@@ -0,0 +1,49 @@
...
+    if show_time:
+        movie_session.show_time = show_time
+    if movie_id:
+        movie_session.movie_id = movie_id
+    if cinema_hall_id:
+        movie_session.cinema_hall_id = cinema_hall_id
+    movie_session.save()
```
MNDonut: still not fixed. Oleksl888 meant that you should get session and save it only when at least 1 parameter is passed, otherwise it doesn't make any sense to fetch session and don't update it because no parameters were passed

MNDonut (changes requested):

db/models.py
```diff
@@ -14,3 +14,39 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="movie_actors")
+    genres = models.ManyToManyField(Genre, related_name="movie_genres")
```
YuriiKindrat, line 23: are u sure about related names ?
this will work like this
`actor.movie_actors.all()` to get all movies related to actor

services/movie.py
```diff
@@ -0,0 +1,36 @@
...
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: Optional[List[int]] = None,
+        actors_ids: Optional[List[int]] = None
+) -> None:
+    movie = Movie.objects.create(
+        title=movie_title,
+        description=movie_description)
```
YuriiKindrat, line 32: bad formatting here

YuriiKindrat (changes requested):

PR #754, opened 2024-04-23
Solution

messedxup:


db/models.py
```diff
@@ -14,3 +14,40 @@ class Actor(models.Model):
...
+    @property
+    def capacity(self) -> float:
+        return self.rows * self.seats_in_row
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall,
+                                    on_delete=models.CASCADE,
+                                    related_name="halls")
```
Oleksl888, line 46: misleading related names. This must show how movie_session object will be called on cinema_hall or movie instances

services/cinema_hall.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+def create_cinema_hall(hall_name: str,
+                       hall_rows: int,
+                       hall_seats_in_row: int) -> CinemaHall:
+    cinema_hall = CinemaHall.objects.create(name=hall_name,
+                                            rows=hall_rows,
+                                            seats_in_row=hall_seats_in_row)
+
+    return cinema_hall
```
Oleksl888, line 17: there is no requirement to return newly created instance

services/movie.py
```diff
@@ -0,0 +1,44 @@
+from django.db.models.query import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(genres_ids: list[int] = None,
```
Oleksl888, line 6: Use Optional for typing here and other functions where appropriate

services/movie.py
```diff
@@ -0,0 +1,44 @@
...
+               actors_ids: list[int] = None) -> QuerySet:
+    result_query = Movie.objects.all()
+
+    if genres_ids and actors_ids:
```
Oleksl888, line 10: this condition is unnecessary since you have the other two

services/movie.py
```diff
@@ -0,0 +1,44 @@
...
+
+    if actors_ids:
+        result_query = result_query.filter(actors__id__in=actors_ids)
+
+    return result_query
+
+
+def get_movie_by_id(movie_id: int) -> Movie | None:
+    if not movie_id:
+        print("Please enter a movie id")
```
Oleksl888, line 25: no need to add this - if there is no id let the error happen

services/movie.py
```diff
@@ -0,0 +1,44 @@
...
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(movie_title: str,
+                 movie_description: str,
+                 genres_ids: list[int] = None,
+                 actors_ids: list[int] = None) -> Movie:
+    movie = Movie.objects.create(title=movie_title,
+                                 description=movie_description)
+    if genres_ids:
```
Oleksl888, line 36: Do as per checklist:
```
Use the .set() method to add a couple of many-to-many relationships (imagine skills is ManyToManyField).

File: services/movie.py [hunk 7]
```diff
@@ -0,0 +1,44 @@
...
+        for genre_id in genres_ids:
+            movie.genres.add(genre_id)
+
+    if actors_ids:
+        for actor_id in actors_ids:
+            movie.actors.add(actor_id)
+
+    return movie
```
Oleksl888, line 44: no need to return a movie here

services/movie_session.py
```diff
@@ -0,0 +1,51 @@
...
+    if cinema_hall_id:
+        movie_session.cinema_hall_id = cinema_hall_id
+    if movie_id:
+        movie_session.movie_id = movie_id
+    movie_session.save()
+    return movie_session
+
+
+def delete_movie_session_by_id(session_id: int) -> QuerySet | None:
+    movie_session = MovieSession.objects.get(id=session_id)
```
Oleksl888, line 48: you can do this in one line.
and no need to return a deleted instance in the end

services/movie_session.py
```diff
@@ -0,0 +1,51 @@
...
+
+def get_movie_session_by_id(movie_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_id)
+
+
+def update_movie_session(session_id: int,
+                         show_time: datetime = None,
+                         movie_id: int = None,
+                         cinema_hall_id: int = None) -> MovieSession:
+    movie_session = MovieSession.objects.get(id=session_id)
```
Oleksl888, line 36: try to rewrite this logic for efficiency, so you don't have to get session and save it in case if you didn't get any optional params. Also returning movie_session is not a requirement

Oleksl888 (changes requested):

services/movie.py
```diff
@@ -0,0 +1,36 @@
...
+        result_query = result_query.filter(genres__id__in=genres_ids)
+
+    if actors_ids:
+        result_query = result_query.filter(actors__id__in=actors_ids)
+
+    return result_query
+
+
+def get_movie_by_id(movie_id: int) -> Optional[Movie]:
+    if not movie_id:
```
Oleksl888, line 22: There is no need for this - here would be fine just to return `Movie.objects.get(id=movie_id)` and if error happens you would catch it wherever you call this service

services/movie.py
```diff
@@ -0,0 +1,36 @@
...
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(movie_title: str,
+                 movie_description: str,
+                 genres_ids: Optional[List[int]] = None,
+                 actors_ids: Optional[List[int]] = None) -> None:
+    movie = Movie.objects.create(title=movie_title,
+                                 description=movie_description)
+    if genres_ids is not None:
```
Oleksl888, line 33: if genres_ids is more appropriate

settings.py
```diff
@@ -21,6 +21,6 @@
 
 USE_I18N = True
 
-USE_TZ = False
+USE_TZ = True
 
-INSTALLED_APPS = ("db",)
+INSTALLED_APPS = ("db", "services")
```
Oleksl888, line 26: services is not really an app in our project

services/movie_session.py
```diff
@@ -0,0 +1,52 @@
...
+from db.models import MovieSession
+
+
+def create_movie_session(movie_show_time: datetime.datetime,
+                         movie_id: int,
+                         cinema_hall_id: int) -> MovieSession:
+    movie_session = MovieSession.objects.create(show_time=movie_show_time,
+                                                movie_id=movie_id,
+                                                cinema_hall_id=cinema_hall_id)
+    return movie_session
```
Oleksl888, line 14: no need to return this

services/movie_session.py
```diff
@@ -0,0 +1,52 @@
...
+def get_movie_session_by_id(movie_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_id)
+
+
+def update_movie_session(session_id: int,
+                         show_time: datetime = None,
+                         movie_id: int = None,
+                         cinema_hall_id: int = None) -> None:
+    try:
+        movie_session = MovieSession.objects.get(id=session_id)
```
Oleksl888, line 37: Here you could do like so:
1. Create an empty dict
2. Fill out the dict one by one In your if statements in case if you received value to update
3. IF you have any values in dict in the end -> update session with update dict

Oleksl888 (changes requested):

services/movie.py
```diff
@@ -0,0 +1,34 @@
...
+        result_query = result_query.filter(genres__id__in=genres_ids)
+
+    if actors_ids:
+        result_query = result_query.filter(actors__id__in=actors_ids)
+
+    return result_query
+
+
+def get_movie_by_id(movie_id: int) -> Optional[Movie]:
+    return Movie.objects.get(id=movie_id)
```
Oleksl888, line 22: Using `pk` is preferable in ORM

services/movie_session.py
```diff
@@ -0,0 +1,52 @@
...
+
+def get_movie_session_by_id(movie_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_id)
+
+
+def update_movie_session(session_id: int,
+                         show_time: datetime = None,
+                         movie_id: int = None,
+                         cinema_hall_id: int = None) -> None:
+    try:
```
Oleksl888, line 34: you don't need to get an object here - all the update logic will be handled on line 48. If the session doesn't exist let the traceback happen - you will try to catch it wherever `update_movie_session` will be called

Oleksl888 (approved):
Well done sir!

PR #756, opened 2024-04-24
'Solution'

dkotkod:


db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="movies_actors")
```
Oleksl888, line 22: Think of a simpler name - it should represent how movies will be accessed from actors or genres instance

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
...
+    def capacity(self) -> int:
+        return self.rows * self.seats_in_row
+
+    def __str__(self) -> str:
+        return f"{self.name}"
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall, on_delete=models.CASCADE)
```
Oleksl888, line 44: related names here?

services/movie.py
```diff
@@ -0,0 +1,41 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: int = None,
```
Oleksl888, line 7: These will be a list of ints. Also use Optional type with optional params

services/movie.py
```diff
@@ -0,0 +1,41 @@
...
+        actors_ids: int = None
+) -> QuerySet:
+    queryset = Movie.objects.all()
+    if genres_ids and actors_ids:
```
Oleksl888, line 11: two if's would be enough here

services/movie.py
```diff
@@ -0,0 +1,41 @@
...
+        )
+    elif genres_ids:
+        queryset = queryset.filter(genres__id__in=genres_ids)
+    elif actors_ids:
+        queryset = queryset.filter(actors__id__in=actors_ids)
+    return queryset.distinct()
+
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
```
Oleksl888, line 24: it is a convention to use `pk` in ORM. `id` is possible but you may have cases when primary key and id are different things

services/movie.py
```diff
@@ -0,0 +1,41 @@
...
+        movie_title: str,
+        movie_description: str,
+        genres_ids: int = None,
+        actors_ids: int = None
+) -> Movie:
+    movie = Movie.objects.create(
+        title=movie_title,
+        description=movie_description
+    )
+    if genres_ids:
```
Oleksl888, line 37:
```
Use the .set() method to add a couple of many-to-many relationships (imagine skills is ManyToManyField).

File: services/movie.py [hunk 7]
```diff
@@ -0,0 +1,41 @@
...
+
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: int = None,
```
Oleksl888, line 30: fix types here too

services/cinema_hall.py
```diff
@@ -0,0 +1,19 @@
...
+def get_cinema_halls() -> QuerySet:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(
+        hall_name: str,
+        hall_rows: int,
+        hall_seats_in_row: int
+) -> CinemaHall:
+    return CinemaHall.objects.create(
```
Oleksl888, line 15: there is no requirement to return an object after creation

services/movie_session.py
```diff
@@ -0,0 +1,48 @@
...
+
+from db.models import MovieSession
+
+
+def create_movie_session(
+        movie_show_time: str,
+        movie_id: int,
+        cinema_hall_id: int
+) -> MovieSession:
+    return MovieSession.objects.create(
```
Oleksl888, line 11: No need to return anything on create/update/delete. Please check in all functions

services/movie_session.py
```diff
@@ -0,0 +1,48 @@
...
+    if movie_id:
+        session.movie_id = movie_id
+    if cinema_hall_id:
+        session.cinema_hall_id = cinema_hall_id
+    session.save()
+    return session
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    session = MovieSession.objects.get(id=session_id)
```
Oleksl888, line 47: you can do it in one line

services/movie_session.py
```diff
@@ -0,0 +1,48 @@
...
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: str = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> MovieSession:
+    session = MovieSession.objects.get(id=session_id)
```
Oleksl888, line 35: try to rewrite this logic for efficiency, so you don't have to get session and save it in case if you didn't get any optional params. Also returning session is not a requirement

Oleksl888 (changes requested):

PR #758, opened 2024-04-24
Solution

siaserb:


db/models.py
```diff
@@ -14,3 +14,36 @@ class Actor(models.Model):
...
+        return self.rows * self.seats_in_row
+
+    def __str__(self) -> str:
+        return self.name
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall, on_delete=models.CASCADE)
+    movie = models.ForeignKey(Movie, on_delete=models.CASCADE)
```
YuriiKindrat, line 45: what about related name here ?

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
...
+
+def create_movie_session(
+        movie_show_time: datetime,
+        movie_id: int,
+        cinema_hall_id: int
+) -> MovieSession:
+    return MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id)
```
YuriiKindrat, line 16: bad formatting here

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
...
+
+def get_movies_sessions(session_date: str = None) -> QuerySet | MovieSession:
+    if session_date:
+        year, month, day = session_date.split("-")
+        return MovieSession.objects.filter(
+            show_time__year=year,
+            show_time__month=month,
+            show_time__day=day
+        )
+    return MovieSession.objects.all()
```
YuriiKindrat, line 27: try to simplify this using `show_time__date` here

YuriiKindrat (changes requested):

services/movie.py
```diff
@@ -0,0 +1,40 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
```
Oleksl888, line 7: Use Optional type hint in cases like these. Check everywhere

services/movie.py
```diff
@@ -0,0 +1,40 @@
...
+        actors_ids: list[int] = None
+) -> QuerySet | Movie:
```
Oleksl888, line 9: does `Movie.objects.all()` also return `Movie` type?

services/movie.py
```diff
@@ -0,0 +1,40 @@
...
+    if genres_ids:
+        movies = movies.filter(genres__id__in=genres_ids)
+    if actors_ids:
+        movies = movies.filter(actors__id__in=actors_ids)
+
+    return movies
+
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
```
Oleksl888, line 22: it is a convention to use `pk` when working with ORM

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+
+from db.models import MovieSession
+
+
+def create_movie_session(
+        movie_show_time: datetime,
+        movie_id: int,
+        cinema_hall_id: int
+) -> MovieSession:
+    return MovieSession.objects.create(
```
Oleksl888, line 13: There is no requirement to return a newly created instance

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet | MovieSession:
```
Oleksl888, line 20: check typing here

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: datetime = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> None:
+    updated_session = MovieSession.objects.get(id=session_id)
```
Oleksl888, line 37: Can you rewrite this so we don't have to get movie session and call save on it in case if we don't have any values to update?

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet | MovieSession:
+    if session_date:
+        session_date = datetime.strptime(session_date, "%Y-%m-%d")
+        return MovieSession.objects.filter(show_time__date=session_date)
```
Oleksl888, line 23: Not a mistake but you can also pass session_date as is to filter and ORM will figure it out

Oleksl888 (changes requested):

YuriiKindrat (approved):
GJ

PR #760, opened 2024-04-25
'Solution'

PozniakVA:


db/migrations/0001_initial.py
```diff
@@ -1,44 +1,58 @@
-# Generated by Django 4.0.2 on 2022-08-19 07:39
```
pavlopro, line 1: Do not change the existing migration file
Add new one instead

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField(null=True, blank=True)
+    actors = models.ManyToManyField(Actor, blank=True)
```
pavlopro, line 22: Add related_name attribute for relational fields

pavlopro (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #764, opened 2024-04-28
Solution

Ruxxell54:


db/models.py
```diff
@@ -14,3 +15,35 @@ class Actor(models.Model):
...
+
+    @property
+    def capacity(self) -> int:
+        return self.rows * self.seats_in_row
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall, on_delete=models.CASCADE)
+    movie = models.ForeignKey(Movie, on_delete=models.CASCADE)
```
YuriiKindrat, line 46: what about related names here ?

services/movie.py
```diff
@@ -0,0 +1,41 @@
...
+from db.models import Movie
+
+
+def get_movies(genres_ids: list[int] = None,
+               actors_ids: list[int] = None) -> QuerySet:
+    queryset = Movie.objects.all()
+
+    if genres_ids is not None and actors_ids is not None:
+        queryset = queryset.filter(genres__id__in=genres_ids,
+                                   actors__id__in=actors_ids)
```
YuriiKindrat, line 12: why do u need this ?

services/movie.py
```diff
@@ -0,0 +1,41 @@
...
+    elif genres_ids is not None:
+        queryset = queryset.filter(genres__id__in=genres_ids)
+    elif actors_ids is not None:
+        queryset = queryset.filter(actors__id__in=actors_ids)
```
YuriiKindrat, line 16: will it work correctly if we will pass the blank list ?

services/movie.py
```diff
@@ -0,0 +1,41 @@
...
+) -> Movie:
+
+    movie = Movie.objects.create(title=movie_title,
+                                 description=movie_description)
+
+    if genres_ids is not None:
+        movie.genres.add(*genres_ids)
+
+    if actors_ids is not None:
+        movie.actors.add(*actors_ids)
```
YuriiKindrat, line 39: same is here

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
+import datetime
+
+from django.db.models.query import QuerySet
+
+from db.models import MovieSession
+
+
+def create_movie_session(
+        movie_show_time: datetime, *,
```
YuriiKindrat, line 9: why do u need * here ?

YuriiKindrat (changes requested):

PR #769, opened 2024-04-29
Develop

bwaah1:


db/models.py
```diff
@@ -14,3 +14,36 @@ class Actor(models.Model):
...
+        return self.name
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall, on_delete=models.PROTECT)
+    movie = models.ForeignKey(Movie, on_delete=models.PROTECT)
+
+    def __str__(self) -> str:
+        return f"{self.movie} {self.show_time}"
```
YuriiKindrat, line 49: what about related names, also why do u use protect option here ?

services/movie.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
+) -> QuerySet[Movie]:
+    if genres_ids and actors_ids:
+        genre_filter = Movie.objects.filter(genres__id__in=genres_ids)
+        actors_ids = genre_filter.filter(actors__id__in=actors_ids)
+        return actors_ids.distinct()
```
YuriiKindrat, line 13: why do u need this ?

YuriiKindrat (changes requested):

YuriiKindrat (approved):
GJ

PR #778, opened 2024-05-06
Solution

VasylRybalka:


services/cinema_hall.py
```diff
@@ -0,0 +1,15 @@
+from django.db.models import QuerySet
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> QuerySet:
```
EdAlekseiev, line 5: You should specify inner type of `QuerySet`
```suggestion
def get_cinema_halls() -> QuerySet[CinemaHall]:

```

services/movie.py
```diff
@@ -0,0 +1,39 @@
+from typing import List, Optional
+from django.db.models import Q, QuerySet
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: Optional[List[int]] = None,
+        actors_ids: Optional[List[int]] = None) -> QuerySet:
```
EdAlekseiev, line 8:
```suggestion
        actors_ids: Optional[List[int]] = None) -> QuerySet[Movie]:

```

services/movie_session.py
```diff
@@ -0,0 +1,51 @@
...
+    new_session = MovieSession(
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
+    new_session.save()
+    return new_session
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
```
EdAlekseiev, line 21: If argument can be `None` you should show it in type hint
```suggestion
def get_movies_sessions(session_date: str | None = None) -> QuerySet[MovieSession]:

```

services/movie_session.py
```diff
@@ -0,0 +1,51 @@
...
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+    session_id: int,
+    show_time: datetime = None,
+    movie_id: int = None,
+    cinema_hall_id: int = None
```
EdAlekseiev, line 36:
```suggestion
    session_id: int,
    show_time: datetime | None = None,
    movie_id: int | None = None,
    cinema_hall_id: int | None = None

```

EdAlekseiev (changes requested):

PR #780, opened 2024-05-09
Solution

lmidzhak:


services/movie.py
```diff
@@ -0,0 +1,30 @@
...
+
+def get_movies(genres_ids: list = None, actors_ids: list = None) -> QuerySet:
+    movie = Movie.objects.all()
+    if genres_ids is not None and actors_ids is not None:
+        movie = movie.filter(genres__id__in=genres_ids,
+                             actors__id__in=actors_ids)
+    if genres_ids is not None:
+        movie = movie.filter(genres__id__in=genres_ids)
+    if actors_ids is not None:
+        movie = movie.filter(actors__id__in=actors_ids)
```
EdAlekseiev, line 13: You can simplify your code here
```suggestion
    if genres_ids is not None:
        movie = movie.filter(genres__id__in=genres_ids)
    if actors_ids is not None:
        movie = movie.filter(actors__id__in=actors_ids)

```

services/movie.py
```diff
@@ -0,0 +1,30 @@
+from db.models import Movie
+from django.db.models import QuerySet
+
+
+def get_movies(genres_ids: list = None, actors_ids: list = None) -> QuerySet:
```
EdAlekseiev, line 5: You should specify inenr type of `list` and `QuerySet`. Also, if argument can be `None` you should show it in type hint. Fix everywhere, please
```suggestion
def get_movies(genres_ids: list[int] | None = None, actors_ids: list[int] | None = None) -> QuerySet[Movie]:

```

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #788, opened 2024-05-16
Solution

o-zinevych:


services/cinema_hall.py
```diff
@@ -0,0 +1,19 @@
+from django.db.models import QuerySet
+
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> QuerySet:
```
EdAlekseiev, line 6: You must show inner type of `QuerySet[CinemaHall]` in type hint. Fix everywhere, please

services/movie.py
```diff
@@ -0,0 +1,35 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
EdAlekseiev, line 8: If argument can be None, you should show it in type hint. Fi everywhere
```suggestion
        genres_ids: list[int] | None = None,
        actors_ids: list[int] | None = None

```

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #789, opened 2024-05-16
Solution

sind14:


services/cinema_hall.py
```diff
@@ -0,0 +1,17 @@
+from django.db.models import QuerySet
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> QuerySet:
```
EdAlekseiev, line 5: You must specify inner type of `QuerySet`. Fix everywhere, please
```suggestion
def get_cinema_halls() -> QuerySet[CinemaHall]:

```

services/movie.py
```diff
@@ -0,0 +1,38 @@
+from django.db.models import QuerySet
+from db.models import Movie
+
+
+def get_movies(genres_ids: list[int] = None,
+               actors_ids: list[int] = None) -> QuerySet:
```
EdAlekseiev, line 6: If argument can be None, you must show it in typehint. Fix everywhere, please
```suggestion
def get_movies(genres_ids: list[int] | None = None,
               actors_ids: list[int] | None = None) -> QuerySet:

```

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #796, opened 2024-06-02
solved

serhii-chaban:


services/cinema_hall.py
```diff
@@ -0,0 +1,22 @@
+from django.db.models import QuerySet
```
Dimosphen1, line 1: Add a blank line between different import groups here and in the other files to follow PEP-8 import rules

services/cinema_hall.py
```diff
@@ -0,0 +1,22 @@
...
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> QuerySet:
```
Dimosphen1, line 5: Extend your return type annotation indicating queryset of what model you return. Apply this change here and in the other places

services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+        show_time=movie_show_time,
+        movie=get_movie_by_id(movie_id),
+        cinema_hall=get_hall_by_id(cinema_hall_id)
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
+    if not session_date:
+        return MovieSession.objects.all()
+    year, month, day = session_date.split("-")
```
Dimosphen1, line 23: There is no need to make split here, pass string object to filtration as is to the Django ORM

Dimosphen1 (changes requested):

db/migrations/0001_initial.py
Dimosphen1, line 1: Do not overwrite the existing migration, create a new one instead.

services/movie.py
```diff
@@ -0,0 +1,44 @@
...
+    if genres_ids and actors_ids:
+        return Movie.objects.all().filter(
+            genres__id__in=genres_ids, actors__id__in=actors_ids
+        )
+
+    if genres_ids and not actors_ids:
+        return Movie.objects.filter(genres__id__in=genres_ids)
+
+    if actors_ids and not genres_ids:
+        return Movie.objects.filter(actors__id__in=actors_ids)
```
Dimosphen1, line 23: Leave only one return statement in this method. You could define an initial `queryset` variable at the beginning, add two if conditions to add filtration to queryset if corresponding arguments are present. Remember that querysets are lazy

services/movie_session.py
```diff
@@ -0,0 +1,51 @@
...
+        movie=get_movie_by_id(movie_id),
+        cinema_hall=get_hall_by_id(cinema_hall_id)
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet[MovieSession]:
+    if not session_date:
+        return MovieSession.objects.all()
+    return MovieSession.objects.filter(
+        show_time__date=session_date)
```
Dimosphen1, line 25: Leave only one return here

Dimosphen1, line 25: In case there are no arguments on the same line with an opening parenthesis, your closing parenthesis should not have any arguments on the same line as well. Please, resolve this issue

Dimosphen1 (changes requested):

PR #801, opened 2024-06-09
Develop

poplyasita:


db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
```
pavlopro, line 22: Please provive related_name attribte for relational fields

services/movie.py
```diff
@@ -0,0 +1,36 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: [int] = None,
+        actors_ids: [int] = None
+) -> QuerySet | Movie:
```
pavlopro, line 9: Please correct return annotation to `QuerySet[Movie]`

services/movie_session.py
```diff
@@ -0,0 +1,48 @@
...
+        movie_session.show_time = show_time
+    if movie_id:
+        movie_session.movie_id = movie_id
+    if cinema_hall_id:
+        movie_session.cinema_hall_id = cinema_hall_id
+    movie_session.save()
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    MovieSession.objects.filter(id=session_id).delete()
```
pavlopro, line 48: Use `get_movie_session_by_id` method

services/movie_session.py
```diff
@@ -0,0 +1,48 @@
...
+    movie_session = MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
+    return movie_session
+
+
+def get_movies_sessions(session_date: str = None) -> [MovieSession]:
+    movie_session = MovieSession.objects.all()
```
pavlopro, line 20: Correct the variable name to `movie_sessions`. 
It would be better to filter by `session_date` to simplify the code, but it's not a big deal.

pavlopro (changes requested):

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="actors")
```
pavlopro, line 22: There aren't correct related names
There should be something like this "movies"

pavlopro (changes requested):

pavlopro (approved):
Good Job!

PR #805, opened 2024-06-12
solution

daikj88:


services/cinema_hall.py
```diff
@@ -0,0 +1,19 @@
+from django.db.models import QuerySet
+
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> QuerySet:
```
Dimosphen1, line 6: Indicate a QuerySet of what instances you return in the return type annotation here and in the other methods as well in the future

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration

PR #812, opened 2024-06-18
solution

AndrewwSSS:


db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
```
DevSpaciX, line 23: It is good practice to add `related_name` field. So pls do it : )

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
...
+        return self.rows * self.seats_in_row
+
+    def __str__(self) -> str:
+        return f"{self.name}"
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall, on_delete=models.DO_NOTHING)
+    movie = models.ForeignKey(Movie, on_delete=models.DO_NOTHING)
```
DevSpaciX, line 45: Why did you chose DO_NOTHING here?

services/movie.py
```diff
@@ -0,0 +1,33 @@
...
+               actors_ids: list[int] = None) -> QuerySet:
+    query = Movie.objects.all()
+
+    if genres_ids and actors_ids:
+        query = query.filter(genres__id__in=genres_ids,
+                             actors__id__in=actors_ids)
+    elif genres_ids:
+        query = query.filter(genres__id__in=genres_ids)
+    elif actors_ids:
+        query = query.filter(actors__id__in=actors_ids)
```
DevSpaciX, line 15: If you delete first condition , nothing will change , cuz your query applies filter on each condition , so you do not need catch first case

DevSpaciX (changes requested):

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="movies")
+    genres = models.ManyToManyField(Genre, related_name="movies")
```
DevSpaciX, line 23: Same related name for different fields that is dangerous . It should be movie_actors or smth

services/movie_session.py
```diff
@@ -0,0 +1,40 @@
...
+def create_movie_session(movie_show_time: datetime,
+                         movie_id: int,
+                         cinema_hall_id: int) -> None:
+    MovieSession.objects.create(movie_id=movie_id,
+                                show_time=movie_show_time,
+                                cinema_hall_id=cinema_hall_id)
+
+
+def get_movies_sessions(session_date: str = None) -> list[MovieSession]:
+    if session_date:
```
DevSpaciX, line 16: It should be `QuerySet[MovieSession]:`
It is not simple list

services/movie.py
```diff
@@ -0,0 +1,31 @@
...
+        query = query.filter(genres__id__in=genres_ids)
+
+    if actors_ids:
+        query = query.filter(actors__id__in=actors_ids)
+
+    return query
+
+
+def get_movie_by_id(movie_id: int) -> object:
+    return Movie.objects.get(id=movie_id)
```
DevSpaciX, line 19: Yeah it is object but , actually it is `Movie`

DevSpaciX (changes requested):

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="movie_actors")
+    genres = models.ManyToManyField(Genre, related_name="movies")
```
DevSpaciX, line 23: Related name should probably be `genres`.

DevSpaciX (changes requested):

services/movie_session.py
```diff
@@ -0,0 +1,42 @@
...
+        session.show_time = show_time
+    if movie_id:
+        session.movie = get_movie_by_id(movie_id)
+    if cinema_hall_id:
+        session.cinema_hall = CinemaHall.objects.get(id=cinema_hall_id)
+    session.save()
+
+
+def delete_movie_session_by_id(id_: int) -> None:
+    MovieSession.objects.get(id=id_).delete()
```
DevSpaciX, line 42: `id_ `it is not very good practice to name viriable : ) Next time use `session_id` or smth

DevSpaciX (approved):

PR #814, opened 2024-06-19
py-movie-session-cinemahall v1

BornToLivee:


db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="movies_actors")
+    genres = models.ManyToManyField(Genre, related_name="movie_genres")
```
DevSpaciX, line 23: why in one case you decided to write `movies` and in the other just a `movie` ?

services/movie_session.py
```diff
@@ -0,0 +1,47 @@
...
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+    session_id: int,
+    show_time: datetime = None,
+    movie_id: int = None,
+    cinema_hall_id: int = None,
+) -> None:
+    movie_session = MovieSession.objects.get(id=session_id)
```
DevSpaciX, line 35: Look , why not to call here your method for this operation ? I am about `get_movie_session_by_id`

services/movie_session.py
```diff
@@ -0,0 +1,47 @@
...
+        movie_session.show_time = show_time
+    if movie_id:
+        movie_session.movie_id = movie_id
+    if cinema_hall_id:
+        movie_session.cinema_hall_id = cinema_hall_id
+    movie_session.save()
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    movie_session = MovieSession.objects.get(id=session_id)
```
DevSpaciX, line 46: Same here , use `get_movie_session_by_id` method : )

db/migrations/0002_cinemahall_movie_moviesession.py
```diff
@@ -0,0 +1,42 @@
...
+            ],
+        ),
+        migrations.CreateModel(
+            name='Movie',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('title', models.CharField(max_length=255)),
+                ('description', models.TextField()),
+                ('actor', models.ManyToManyField(related_name='movies_actors', to='db.Actor')),
+                ('genres', models.ManyToManyField(related_name='movie_genres', to='db.Genre')),
```
DevSpaciX, line 30: Same here with related name . Use one variant in singular or plural form

DevSpaciX (changes requested):

DevSpaciX (approved):
Great !

PR #815, opened 2024-06-22
created cinema db with funcs

Timur5050:


services/movie_session.py
```diff
@@ -0,0 +1,47 @@
...
+    if movie_id:
+        movie_to_change.movie_id = movie_id
+    if cinema_hall_id:
+        movie_to_change.cinema_hall_id = cinema_hall_id
+
+    movie_to_change.save()
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    get_movie_session_by_id(session_id).delete()
```
vsmutok, line 47: use `get_movies_sessions` here

vsmutok (changes requested):

Timur5050:
i had like that and vsmutok said me to use there get_movies_sessions, u can read that in previous message. i dont understand what problems

PR #816, opened 2024-06-22
Develop

jpdreamthug:


db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="movies")
+    genres = models.ManyToManyField(Genre, related_name="movies")
```
DevSpaciX, line 23: Same related name on a different fields ?? That is dangerous . It should be `genres` or  `actors` accordingly

services/movie.py
```diff
@@ -0,0 +1,51 @@
...
+
+    return query.distinct()
+
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    try:
+        movie = Movie.objects.get(id=movie_id)
+        return movie
+    except Movie.DoesNotExist:
+        print("Enter valid Movie id")
```
DevSpaciX, line 31: If you want to catch the case if Movie does not exist by some id . Use `get_object_or_404` function which comes from Django by default 
`from django.shortcuts import get_object_or_404
`

services/movie_session.py
```diff
@@ -0,0 +1,58 @@
...
+        )
+
+    return sessions
+
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    try:
+        return MovieSession.objects.get(id=movie_session_id)
+    except MovieSession.DoesNotExist:
+        print("Enter valid Movie session id")
```
DevSpaciX, line 35: Same here )

services/movie_session.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: str | None = None,
+        movie_id: int | None = None,
+        cinema_hall_id: int | None = None
+) -> MovieSession:
+
+    movie = MovieSession.objects.filter(id=session_id)
```
DevSpaciX, line 45: Use your `get_movie_session_by_id` method

services/movie_session.py
```diff
@@ -0,0 +1,58 @@
...
+    if movie_id:
+        movie.update(movie_id=movie_id)
+    if cinema_hall_id:
+        movie.update(cinema_hall_id=cinema_hall_id)
+
+    return movie
+
+
+def delete_movie_session_by_id(session_id: int) -> MovieSession:
+    return MovieSession.objects.filter(id=session_id).delete()
```
DevSpaciX, line 58: If you want to get some data using id . Do not use `filter()` , use `get()` instead !

DevSpaciX (changes requested):

DevSpaciX (approved):
Good job!

PR #817, opened 2024-06-22
Solution

BohdanKuzik:


services/movie_session.py
```diff
@@ -0,0 +1,56 @@
...
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: str = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> MovieSession:
+    movie_session = MovieSession.objects.get(id=session_id)
```
DevSpaciX, line 42: Look , why not to call here your method for this operation ? I am about `get_movie_session_by_id`

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
```
DevSpaciX, line 23: It is good practice to add `related_name` field. So pls do it : )

DevSpaciX (changes requested):

services/movie_session.py
```diff
@@ -0,0 +1,56 @@
...
+        show_time: str = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> MovieSession:
+    movie_session = get_movie_session_by_id(session_id)
+
+    if show_time:
+        movie_session.show_time = show_time
+    if movie_id:
+        movie_session.movie = Movie.objects.get(id=movie_id)
```
DevSpaciX, line 47: Maybe better use  `get_movie_by_id` func ?)

DevSpaciX (changes requested):

PR #819, opened 2024-06-24
sol1

Unlie9:


db/models.py
```diff
@@ -14,3 +14,36 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
```
DevSpaciX, line 23: Add `related_fields`

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+def create_movie_session(movie_show_time: Movie, movie_id: Movie,
+                         cinema_hall_id: CinemaHall) -> None:
+    movie = Movie.objects.get(id=movie_id)
+    cinema_hall = CinemaHall.objects.get(id=cinema_hall_id)
+
+    MovieSession.objects.create(
+        show_time=movie_show_time,
+        cinema_hall=cinema_hall,
+        movie=movie
+    )
```
DevSpaciX, line 15: You could set here just ids . For example 
`cinema_hall_id=cinema_hall_id`

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+def get_movie_session_by_id(movie_session_id: Movie) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(session_id: MovieSession,
+                         show_time: MovieSession = None,
+                         movie_id: Movie = None,
+                         cinema_hall_id: CinemaHall = None) -> MovieSession:
+    session = MovieSession.objects.get(id=session_id)
+
```
DevSpaciX, line 35: Could you use here `get_movie_session_by_id` func ?

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+    if show_time:
+        session.show_time = show_time
+    if movie_id:
+        movie = Movie.objects.get(id=movie_id)
+        session.movie = movie
```
DevSpaciX, line 40: Reuse `get_movie_by_id` func

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+    if cinema_hall_id:
+        cinema_hall = CinemaHall.objects.get(id=cinema_hall_id)
+        session.cinema_hall = cinema_hall
+
+    session.save()
+    return session
+
+
+def delete_movie_session_by_id(session_id: MovieSession) -> MovieSession:
+    MovieSession.objects.filter(id=session_id).delete()
```
DevSpaciX, line 50: If you getting some value by id - use get() 
In this case just use `get_movie_session_by_id`

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #821, opened 2024-06-24
Solution_v1

BileichukIvan:


services/movie_session.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: str = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> MovieSession:
+    movie_session = MovieSession.objects.get(id=session_id)
+    if show_time:
```
DevSpaciX, line 36: Could you just reuse your `get_movie_session_by_id` for this operation ?

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #822, opened 2024-06-24
Solution

lilarin:


services/movie_session.py
```diff
@@ -0,0 +1,56 @@
...
+        movie_session.cinema_hall_id = cinema_hall_id
+    return movie_session.save()
+
+
+def delete_movie_session_by_id(
+        session_id: int
+) -> MovieSession:
+    deleted_session = MovieSession.objects.get(id=session_id)
+    deleted_session.delete()
+    return deleted_session
```
DevSpaciX, line 56: You could make it in one line without returns . 
That is for future : )
`MovieSession.objects.get(id=session_id).delete()`

DevSpaciX (approved):
Check the comment pls )

PR #823, opened 2024-06-24
Solution

VShveda:


services/cinema_hall.py
```diff
@@ -0,0 +1,18 @@
...
+def create_cinema_hall(
+        hall_name: str,
+        hall_rows: int,
+        hall_seats_in_row: int
+) -> QuerySet:
+    return CinemaHall.objects.create(
+        name=hall_name,
+        rows=hall_rows,
+        seats_in_row=hall_seats_in_row
+    )
```
DevSpaciX, line 18: Remember:`.all()`, `.filter()` returns QuerySet`

Fix it everywhere

services/movie.py
```diff
@@ -0,0 +1,37 @@
...
+    movie = Movie.objects.create(
+        title=movie_title,
+        description=movie_description,
+    )
+    if genres_ids:
+        for genre_id in genres_ids:
+            movie.genres.add(genre_id)
+    if actors_ids:
+        for actor_id in actors_ids:
+            movie.actors.add(actor_id)
```
DevSpaciX, line 36: Try to optimize it here using `set()` instead of `add()` and for loop

services/movie_session.py
```diff
@@ -0,0 +1,45 @@
...
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: str = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> MovieSession:
+    movie_sessions = MovieSession.objects.get(id=session_id)
```
DevSpaciX, line 34: Could you use here `get_movie_session_by_id` func ?

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #824, opened 2024-06-25
'Solution'

dsokolovskiy:


db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="actor_movies")
+    genres = models.ManyToManyField(Genre, related_name="genre_movies")
```
DevSpaciX, line 23: Related names here probably should be just a plural form of a "movie". What do you think?

services/cinema_hall.py
```diff
@@ -0,0 +1,18 @@
...
+
+def get_cinema_halls() -> QuerySet:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(
+        hall_name: str,
+        hall_rows: int,
+        hall_seats_in_row: int
+) -> QuerySet:
```
DevSpaciX, line 13: Remember:`.all()`, `.filter()` returns QuerySet`.get()`, `.delete()`, `.create()` returns model class, `-> User` for. example . 
Fix it everywhere if needed

services/movie.py
```diff
@@ -0,0 +1,37 @@
...
+    movie = Movie.objects.create(
+        title=movie_title,
+        description=movie_description,
+    )
+    if genres_ids:
+        for genre_id in genres_ids:
+            movie.genres.add(genre_id)
+    if actors_ids:
+        for actor_id in actors_ids:
+            movie.actors.add(actor_id)
```
DevSpaciX, line 36: Try to optimize it using `set()`

DevSpaciX (changes requested):

DevSpaciX (changes requested):

services/movie.py
```diff
@@ -0,0 +1,36 @@
...
+) -> QuerySet:
+    movie = Movie.objects.create(
+        title=movie_title,
+        description=movie_description,
+    )
+    if genres_ids:
+        movie.genres.add(*set(genres_ids))
+
+    if actors_ids:
+        movie.actors.add(*set(actors_ids))
```
DevSpaciX, line 35: I mean you could do it lke this ->
`movie.actors.set(actors_ids)`

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #825, opened 2024-06-25
solution1

NikitaNikolaitsev:


services/movie_session.py
```diff
@@ -0,0 +1,50 @@
+from django.db.models import QuerySet
+
+from db.models import MovieSession
+
+
+def create_movie_session(
+        movie_show_time: str,
+        movie_id: int,
+        cinematically_id: int
+) -> QuerySet:
```
DevSpaciX, line 10: Remember:`.all()`, `.filter()` returns QuerySet`

Fix it everywhere

DevSpaciX (changes requested):
The tests have not passed fix it pls

DevSpaciX (approved):

PR #829, opened 2024-06-25
Solution

Andrii1000:


db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
```
DevSpaciX, line 23: Do not forget to add `related_names`

services/movie_session.py
```diff
@@ -0,0 +1,46 @@
...
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: str = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None,
+) -> None:
+    movie_session = MovieSession.objects.get(id=session_id)
```
DevSpaciX, line 35: You can use here your method `get_movie_session_by_id`

DevSpaciX (changes requested):

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #830, opened 2024-06-25
Solution

necobra:


services/movie.py
```diff
@@ -0,0 +1,37 @@
...
+        genres_ids: list[int] = None, actors_ids: list[int] = None
+) -> None:
+    movie = Movie.objects.create(
+        title=movie_title,
+        description=movie_description
+    )
+    if genres_ids:
+        movie.genres.add(
+            *Genre.objects.filter(id__in=genres_ids).all()
+        )
```
DevSpaciX, line 33: You can use here `set()` method

services/movie_session.py
```diff
@@ -0,0 +1,52 @@
...
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
+    movies = MovieSession.objects
+    if session_date:
+        movies = movies.filter(
+            show_time__date=datetime.strptime(session_date, "%Y-%m-%d")
+        )
+    return movies.all()
```
DevSpaciX, line 26: Why did you decide to cal .`all()` at the end ? And actually you do not need to override your viriable . You can just filter your existing query : )

services/movie_session.py
```diff
@@ -0,0 +1,52 @@
...
+    return MovieSession.objects.get(id=movies_session_id)
+
+
+def update_movie_session(
+        session_id: int, show_time: datetime = None,
+        movie_id: int = None, cinema_hall_id: int = None
+) -> None:
+    movie_session = MovieSession.objects.filter(
+        id=session_id
+    )
```
DevSpaciX, line 39: For getting data by id using method `get()` is more suitable . And could u reuse your method `get_movie_session_by_id` here ?))

DevSpaciX (changes requested):

db/models.py
```diff
@@ -14,3 +14,41 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
```
DevSpaciX, line 23: Add pls related names here )

services/movie.py
```diff
@@ -0,0 +1,37 @@
...
+        genres_ids: list[int] = None, actors_ids: list[int] = None
+) -> None:
+    movie = Movie.objects.create(
+        title=movie_title,
+        description=movie_description
+    )
+    if genres_ids:
+        movie.genres.set(
+            Genre.objects.filter(id__in=genres_ids)
+        )
```
DevSpaciX, line 33: `movie.genres.set(genres_ids)` try this )

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #832, opened 2024-06-25
solution

DimaYankovskiy:


services/movie_session.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: str = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> MovieSession:
+    session_to_update = MovieSession.objects.get(id=session_id)
+    if show_time:
```
DevSpaciX, line 35: You have already implemented method for this operation ( `get_movie_session_by_id` )

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #833, opened 2024-06-25
Solution

b3v3kt0r:


services/movie.py
```diff
@@ -0,0 +1,32 @@
...
+
+from db.models import Movie
+
+
+def get_movies(genres_ids: list[int] = None,
+               actors_ids: list[int] = None) -> QuerySet | Movie:
+    queryset = Movie.objects.all()
+    if genres_ids and actors_ids:
+        queryset = queryset.filter(genres__id__in=genres_ids)
+        queryset = queryset.filter(actors__id__in=actors_ids)
```
DevSpaciX, line 11: You don’t need this check since your code will go into the two conditions below in this case and filter your queryset in the same way

services/movie.py
```diff
@@ -0,0 +1,32 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(genres_ids: list[int] = None,
+               actors_ids: list[int] = None) -> QuerySet | Movie:
```
DevSpaciX, line 7: Return type hint here will be `QuerySet` 

services/movie_session.py
```diff
@@ -0,0 +1,38 @@
...
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(session_id: int,
+                         show_time: int = None,
+                         movie_id: int = None,
+                         cinema_hall_id: int = None) -> None:
+    session_to_update = MovieSession.objects.filter(id=session_id)
```
DevSpaciX, line 28: If you need to get some value by id ( use .get() ) ALso you already implemented func for this `get_movie_session_by_id` so you can just call it

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #834, opened 2024-06-25
sol1

IvanVahanov:


services/movie.py
```diff
@@ -0,0 +1,33 @@
+from db.models import Movie, Genre, Actor
+
+
+def get_movies(genres_ids: list[int] = None,
+               actors_ids: list[int] = None) -> list[Movie]:
+    movies = Movie.objects.all()
+    if genres_ids:
+        movies = movies.filter(genres__id__in=genres_ids).distinct()
+    if actors_ids:
+        movies = movies.filter(actors__id__in=actors_ids).distinct()
```
DevSpaciX, line 10: Why did you use distinct here ?

services/movie_session.py
```diff
@@ -0,0 +1,39 @@
...
+def create_movie_session(
+        movie_show_time: datetime,
+        movie_id: int, cinema_hall_id: int) -> MovieSession:
+    return MovieSession.objects.create(show_time=movie_show_time,
+                                       movie_id=movie_id,
+                                       cinema_hall_id=cinema_hall_id)
+
+
+def get_movies_sessions(session_date: str = None) -> list[MovieSession]:
+    if session_date:
```
DevSpaciX, line 14: Return type hint here will be `QuerySet` 

services/movie.py
```diff
@@ -0,0 +1,33 @@
+from db.models import Movie, Genre, Actor
+
+
+def get_movies(genres_ids: list[int] = None,
+               actors_ids: list[int] = None) -> list[Movie]:
```
DevSpaciX, line 5: that is not just a list of Movies , that is QuerySet

services/movie_session.py
```diff
@@ -0,0 +1,39 @@
...
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(session_id: int,
+                         show_time: datetime = None,
+                         movie_id: int = None,
+                         cinema_hall_id: int = None) -> MovieSession:
+    session = MovieSession.objects.get(id=session_id)
```
DevSpaciX, line 27: Why not to use here your func `get_movie_session_by_id` ?

DevSpaciX (changes requested):

PR #835, opened 2024-06-25
"Solution"

MaglorFeanorson:


services/cinema_hall.py
```diff
@@ -0,0 +1,18 @@
+from db.models import CinemaHall
+from typing import List
+
+
+def get_cinema_halls() -> List[CinemaHall]:
+    return CinemaHall.objects.all()
```
DevSpaciX, line 6: Return type hint here will be `QuerySet` 

services/movie.py
```diff
@@ -0,0 +1,35 @@
+from db.models import Movie
+from typing import List, Optional
+
+
+def get_movies(
+        genres_ids: Optional[List[int]] = None,
+        actors_ids: Optional[List[int]] = None
+) -> List[Movie]:
```
DevSpaciX, line 8: Same with return type here . Fix it everywhere

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+        movie_show_time: datetime.datetime,
+        movie_id: int,
+        cinema_hall_id: int
+) -> MovieSession:
+    movie = Movie.objects.get(id=movie_id)
+    cinema_hall = CinemaHall.objects.get(id=cinema_hall_id)
+    return MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie=movie,
+        cinema_hall=cinema_hall
```
DevSpaciX, line 16: You could just set `cinema_hall_id`= `cinema_hall_id` . Same with movie . Instead of set whole instance of movie or cinema_hall

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: Optional[datetime.datetime] = None,
+        movie_id: Optional[int] = None,
+        cinema_hall_id: Optional[int] = None
+) -> MovieSession:
+    movie_session = MovieSession.objects.get(id=session_id)
+    if show_time:
```
DevSpaciX, line 39: You already have a func `get_movie_session_by_id` just use it

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+        movie_session.show_time = show_time
+    if movie_id:
+        movie_session.movie = Movie.objects.get(id=movie_id)
```
DevSpaciX, line 42: Same here

DevSpaciX (changes requested):

services/movie_session.py
```diff
@@ -0,0 +1,48 @@
...
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
+
+
+def get_movies_sessions(
+        session_date: Optional[str] = None
+) -> List[MovieSession]:
+    if session_date:
```
DevSpaciX, line 21: Maybe `QuerySet` return type should be ?

services/movie_session.py
```diff
@@ -0,0 +1,48 @@
...
+        movie_id: Optional[int] = None,
+        cinema_hall_id: Optional[int] = None
+) -> MovieSession:
+    movie_session = get_movie_session_by_id(session_id)
+    if show_time:
+        movie_session.show_time = show_time
+    if movie_id:
+        movie_session.movie_id = movie_id
+    if cinema_hall_id:
+        movie_session.cinema_hall = CinemaHall.objects.get(id=cinema_hall_id)
```
DevSpaciX, line 42: Why did you decide to set movie by id and cinema_hall by whole object ? Chose smth one ) 
Imho setting by id it is more clean so ->
`movie_sessions.cinema_hall_id = cinema_hall_id`

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #836, opened 2024-06-26
Solution

rinarice:


services/cinema_hall.py
```diff
@@ -0,0 +1,19 @@
+from django.db.models import QuerySet
+
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> QuerySet:
```
EdAlekseiev, line 6: You should specify inner type of QuerySet. Fix evrywhere, please
```suggestion
def get_cinema_halls() -> QuerySet[CinemaHall]:
```

services/movie.py
```diff
@@ -0,0 +1,38 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None,
```
```suggestion
        genres_ids: list[int] | None = None,
        actors_ids: list[int] | None = None,
```

EdAlekseiev (changes requested):

vasylhnatiuk (approved):

PR #865, opened 2024-07-22
implemented cinema task

MasakDirt:


services/movie.py
```diff
@@ -0,0 +1,45 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
MNDonut, line 8: list or None type

services/movie.py
```diff
@@ -0,0 +1,45 @@
...
+                .filter(actors__id__in=actors_ids)
+                .filter(genres__id__in=genres_ids))
+
+    if genres_ids:
+        return Movie.objects.filter(genres__id__in=genres_ids)
+
+    if actors_ids:
+        return Movie.objects.filter(actors__id__in=actors_ids)
+
+    return Movie.objects.all()
```
MNDonut, line 21: instead of making return everywhere you can create a variable queryset that initially contains all movies.
then you add ifs that filters this queryset(and updates it but doesn't return queryset). In this way python'll reach each if statement and will return what you need

services/movie.py
```diff
@@ -0,0 +1,45 @@
...
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None,
```
MNDonut, line 32: list or None type

services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet[MovieSession]:
+    if session_date:
+        return MovieSession.objects.filter(
+            show_time__date=datetime.strptime(session_date, "%Y-%m-%d"))
```
MNDonut, line 23: bad formatting

services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+
+    if cinema_hall_id != 0:
+        update_fields["cinema_hall_id"] = cinema_hall_id
+
+    if update_fields:
+        MovieSession.objects.filter(id=session_id).update(**update_fields)
+
+
+def delete_movie_session_by_id(movie_session_id: int) -> None:
+    get_movie_session_by_id(movie_session_id).delete()
```
MNDonut, line 54: just a tip for the future that .get() may raise error if you pass id that doesn't exist and delete won't work. you can use .filter().delete() and error will not be raised if filter returns an empty queryset

MNDonut (changes requested):

services/movie_session.py
MasakDirt: Welcome back🎉! Everyone was waiting for you)

db/models.py
```diff
@@ -14,3 +14,36 @@ class Actor(models.Model):
...
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall, on_delete=models.CASCADE)
+    movie = models.ForeignKey(Movie, on_delete=models.CASCADE)
+
+    def __str__(self) -> str:
+        formatted_time = self.show_time.strftime("%Y-%m-%d %H:%M:%S")
+        return f"{self.movie.title} {formatted_time}"
```
Oleksl888, line 49: You can actually make it simple here like so `{str(self.show_time)}` or pass formatted time string without storing it in a variable

services/movie.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def get_movies(
+        genres_ids: list[int] | None = None,
+        actors_ids: list[int] | None = None
+) -> QuerySet[Movie]:
+    movies = Movie.objects.all()
+
+    if genres_ids:
+        movies = movies.filter(genres__id__in=genres_ids)
```
Oleksl888, line 13: `genres__in=genres_ids` will also work

services/movie.py
```diff
@@ -0,0 +1,42 @@
...
+
+    if actors_ids:
+        movies = movies.filter(actors__id__in=actors_ids)
+
+    return movies
+
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
```
Oleksl888, line 22: using `pk` is prererable over `id `

services/movie_session.py
```diff
@@ -0,0 +1,56 @@
...
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet[MovieSession]:
+    movie_sessions = MovieSession.objects.all()
+    if session_date:
+        movie_sessions = movie_sessions.filter(
+            show_time__date=datetime.strptime(session_date, "%Y-%m-%d")
```
Oleksl888, line 24: Or you could more simply filter like so `show_time__date=session_date` django will take care about type conversion

services/movie_session.py
```diff
@@ -0,0 +1,56 @@
...
+        )
+
+    return movie_sessions
+
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
```
Oleksl888, line 31: `pk`

services/movie_session.py
```diff
@@ -0,0 +1,56 @@
...
+        update_fields["show_time"] = show_time
+
+    if movie_id != 0:
+        update_fields["movie_id"] = movie_id
+
+    if cinema_hall_id != 0:
+        update_fields["cinema_hall_id"] = cinema_hall_id
+
+    if update_fields:
+        MovieSession.objects.filter(id=session_id).update(**update_fields)
```
Oleksl888, line 52: I like how you did it here - very nice pattern with update dict

services/movie_session.py
```diff
@@ -0,0 +1,56 @@
...
+        show_time: datetime = None,
+        movie_id: int = 0,
+        cinema_hall_id: int = 0,
+) -> None:
+    update_fields = {}
+
+    if show_time:
+        update_fields["show_time"] = show_time
+
+    if movie_id != 0:
```
Oleksl888, line 45: here can be just `if `movie_id`

services/movie_session.py
```diff
@@ -0,0 +1,56 @@
...
+
+    if cinema_hall_id != 0:
+        update_fields["cinema_hall_id"] = cinema_hall_id
+
+    if update_fields:
+        MovieSession.objects.filter(id=session_id).update(**update_fields)
+
+
+def delete_movie_session_by_id(movie_session_id: int) -> None:
+    MovieSession.objects.filter(id=movie_session_id).delete()
```
Oleksl888, line 56: Here you could call previously defined function to make code DRY ` get_movie_session_by_id`

Oleksl888 (changes requested):

MNDonut (approved):

PR #871, opened 2024-07-29
movie-moviesession-cinemahall sol

TenbReala:


db/migrations/0001_initial.py
Dimosphen1, line 1: Do not adjust the initial migration, it should be left untouched. Create a new migration instead

services/cinema_hall.py
```diff
@@ -0,0 +1,19 @@
+from django.db.models import QuerySet
+
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> QuerySet:
```
Dimosphen1, line 6: Indicate a QuerySet of what instances you return in the return type annotation here and in the other places as well

services/movie_session.py
```diff
@@ -0,0 +1,45 @@
...
+    if movie_id:
+        queryset.movie_id = movie_id
+    if cinema_hall_id:
+        queryset.cinema_hall_id = cinema_hall_id
+    queryset.save()
+    return queryset
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    MovieSession.objects.get(id=session_id).delete()
```
Dimosphen1, line 45: Use `get_movie_session_by_id` function here

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
```
Dimosphen1, line 23: Add `related_name` attribute to all corresponding fields

Dimosphen1 (changes requested):

db/migrations/0001_initial.py
TenbReala: How to fix it?

Polyakiv-Andrey (approved):
Good job!

PR #880, opened 2024-08-13
"Solution"

OleksiukStepan:


db/models.py
```diff
@@ -14,3 +15,38 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField(blank=True, null=True)
```
MNDonut, line 22: it isn't mentioned it can be null or blank

db/models.py
```diff
@@ -1,10 +1,11 @@
 from django.db import models
+from django.db.models import CharField
 
 
 class Genre(models.Model):
     name = models.CharField(max_length=255, unique=True)
 
-    def __str__(self) -> str:
+    def __str__(self) -> CharField:
```
MNDonut, line 8: str is enough
also why to not use models.CharField but import it separately? Because otherwise you should follow the same approach everywhere and use CharField instead of models.CharField that isn't actually good

db/models.py
```diff
@@ -14,3 +15,38 @@ class Actor(models.Model):
...
+    @property
+    def capacity(self) -> int:
+        return self.rows * self.seats_in_row
+
+    def __str__(self) -> str:
+        return self.name
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField(auto_created=True)
```
MNDonut, line 44: I see this parameter for the first time to be honest. I think it shouldn't be auto created because show_time is something that some cinema manager sets manually, so it can't be for example now time

services/movie.py
```diff
@@ -0,0 +1,38 @@
+from django.db.models import QuerySet
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list = None,
+        actors_ids: list = None
```
MNDonut, line 7: list or None type

services/movie.py
```diff
@@ -0,0 +1,38 @@
...
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: list = None,
+        actors_ids: list = None
```
MNDonut, line 27: list or None type

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
...
+        movie_session.movie_id = movie_id
+    if cinema_hall_id:
+        movie_session.cinema_hall_id = cinema_hall_id
+
+    movie_session.save()
+    return movie_session
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    get_movie_session_by_id(session_id).delete()
```
MNDonut, line 53: just a tip for the future that .get() may raise error if you pass id that doesn't exist and delete won't work. you can use .filter().delete() and error will not be raised if filter returns an empty queryset

MNDonut (changes requested):

db/models.py
OleksiukStepan: It only works without parameters.

MNDonut (approved):

PR #882, opened 2024-08-14
develop

BGBroth:


LLkaia (changes requested):
It's a bad idea to replace existed migrations with new one. If another developer has this repository, and then will pull new changes, migrations will not work correctly on his side as well as during ci/cd on production. It's better to create new one instead.

PR #891, opened 2024-08-20
Solution

DSahalatyi:


services/movie.py
```diff
@@ -0,0 +1,52 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+from typing import Optional
+
+
+def get_movies(
+        genres_ids: Optional[list[int]] = None,
+        actors_ids: Optional[list[int]] = None,
```
MNDonut, line 9: it's better to use | operator

services/movie.py
```diff
@@ -0,0 +1,52 @@
...
+
+    if genres_ids:
+        queryset = queryset.filter(genres__id__in=genres_ids)
+        return queryset
+
+    if actors_ids:
+        queryset = queryset.filter(actors__id__in=actors_ids)
+        return queryset
+
+    return queryset
```
MNDonut, line 28: instead of making return everywhere you can do filtering and return it at the end.
in this case you'll not need the first if statement

services/movie_session.py
```diff
@@ -0,0 +1,55 @@
...
+        session.update(movie_id=movie_id)
+
+    if cinema_hall_id:
+        session.update(cinema_hall_id=cinema_hall_id)
+
+    return session
+
+
+def delete_movie_session_by_id(session_id: int) -> MovieSession:
+    session = MovieSession.objects.get(id=session_id).delete()
```
MNDonut, line 54: just a tip for the future that .get() may raise error if you pass id that doesn't exist and delete won't work. you can use .filter().delete() and error will not be raised if filter returns an empty queryset

MNDonut (changes requested):

services/movie.py
```diff
@@ -0,0 +1,49 @@
...
+        genres_ids: list[int] | None = None,
+        actors_ids: list[int] | None = None,
+) -> QuerySet:
+    queryset = Movie.objects.all()
+
+    if genres_ids and actors_ids:
+        queryset = queryset.filter(
+            genres__id__in=genres_ids,
+            actors__id__in=actors_ids
+        )
```
MNDonut, line 17: you don't need it because if both params are passed, then both if statement below will be activated

services/movie.py
```diff
@@ -0,0 +1,49 @@
...
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: Optional[list[int]] = None,
+        actors_ids: Optional[list[int]] = None
```
MNDonut, line 36: let's follow the same code styling everywere and use | operator

services/movie_session.py
```diff
@@ -0,0 +1,55 @@
...
+        cinema_hall_id: int,
+) -> MovieSession:
+    return MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
+
+
+def get_movies_sessions(session_date: Optional[str] = None) -> QuerySet:
```
MNDonut, line 20: | operator

services/movie_session.py
```diff
@@ -0,0 +1,55 @@
...
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: Optional[datetime] = None,
+        movie_id: Optional[int] = None,
+        cinema_hall_id: Optional[int] = None
```
MNDonut, line 37: | operator

MNDonut (changes requested):

MNDonut (approved):

PR #899, opened 2024-08-22
solution

BastovOleksandr:


db/models.py
```diff
@@ -14,3 +18,45 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255, unique=True)
```
MNDonut, line 24: it shouldn't be unique

db/models.py
```diff
@@ -14,3 +18,45 @@ class Actor(models.Model):
...
+    description = models.TextField(null=True, blank=True)
```
MNDonut, line 25: it shoudn't be blank or null

services/movie.py
```diff
@@ -0,0 +1,40 @@
+from typing import Optional
+
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: Optional[list[int]] = None,
+        actors_ids: Optional[list[int]] = None
```
MNDonut, line 10: list[int] | None

services/movie.py
```diff
@@ -0,0 +1,40 @@
...
+) -> QuerySet:
+    movies = Movie.objects.all()
+
+    if genres_ids:
+        movies = movies.filter(genres__id__in=genres_ids).distinct()
+
+    if actors_ids:
+        movies = movies.filter(actors__id__in=actors_ids).distinct()
```
MNDonut, line 18: you can just do distinct() at the end

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
...
+    if movie_id:
+        movie_session.movie_id = movie_id
+    if cinema_hall_id:
+        movie_session.cinema_hall_id = cinema_hall_id
+
+    movie_session.save()
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    MovieSession.objects.get(id=session_id).delete()
```
MNDonut, line 53: just a tip for the future: in case movie is not found, it'll raise error that object does not exist and won't be able to delete
If you used .filter().delete(), then error won't be raised even if the object won't be found

MNDonut (changes requested):

MNDonut (approved):

PR #904, opened 2024-08-24
Solution

Morphin20th:


db/migrations/0003_alter_movie_actors_alter_movie_genres_and_more.py
Dimosphen1, line 1: It is a good practice to include one migration per app per PR in Django projects since it helps to avoid `migrations` folder overload in the future. Represent all changes done to models in one migration file per app per PR, please

db/models.py
```diff
@@ -14,3 +14,42 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="actors")
+    genres = models.ManyToManyField(Genre, related_name="genres")
```
Dimosphen1, line 23: `related_name` is incorrect for both fields. In most cases, the plural form of a model would be applicable (e.g. `"movies"`). Change `related_name` for corresponding `MovieSession` fields as well

services/movie.py
```diff
@@ -0,0 +1,45 @@
...
+        query = query.filter(
+            genres__id__in=genres_ids,
+            actors__id__in=actors_ids
+        )
+
+    if genres_ids and not actors_ids:
+        query = query.filter(genres__id__in=genres_ids)
+
+    if actors_ids and not genres_ids:
+        query = query.filter(actors__id__in=actors_ids)
```
Dimosphen1, line 22: Leave only two if conditions and append new `filter` methods to queryset instead of looking for all combinations. In case there is greater number of arguments, there would be a progressing increase in the number of if conditions

services/movie_session.py
```diff
@@ -0,0 +1,52 @@
...
+        session.show_time = show_time
+    if movie_id:
+        session.movie = Movie.objects.get(id=movie_id)
+    if cinema_hall_id:
+        session.cinema_hall = CinemaHall.objects.get(id=cinema_hall_id)
+    session.save()
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    MovieSession.objects.get(id=session_id).delete()
```
Dimosphen1, line 52: Use `get_movie_session_by_id` here

services/movie_session.py
```diff
@@ -0,0 +1,52 @@
+from datetime import datetime
```
Dimosphen1, line 1: Add a blank line between different groups of imports to follow PEP-8

Dimosphen1 (changes requested):

db/migrations/0003_alter_movie_actors_alter_movie_genres_and_more.py
Morphin20th: sorry, didn't understand what I need to do? Delete all migration files except the last one?

Dimosphen1: No, make sure that you add only one migration that represents all changes done to the models, do not edit or delete the migration that is already present in the project

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
...
+def create_movie_session(
+        movie_show_time: datetime,
+        movie_id: int,
+        cinema_hall_id: int,
+) -> MovieSession:
+
+    movie = Movie.objects.get(id=movie_id)
+    cinema_hall = CinemaHall.objects.get(id=cinema_hall_id)
+
+    movie_session = movie.movie_sessions.create(
```
Dimosphen1, line 17: Use `MovieSession` model to create instance. Besides that, avoid using lines 14 and 15, write ids directly while creating movie session instance

db/migrations/0004_alter_movie_actors_alter_movie_genres_and_more.py
Dimosphen1, line 1: Include only one migration in the PR (pull request)

Dimosphen1 (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #906, opened 2024-08-25
solution

HalynaRohatska:


services/movie.py
```diff
@@ -0,0 +1,43 @@
...
+        actors_ids: list[int] = None,
+) -> Movie:
+    movie = Movie.objects.create(
+        title=movie_title,
+        description=movie_description
+    )
+
+    if genres_ids:
+        genres = Genre.objects.filter(id__in=genres_ids)
+        movie.genres.set(genres)
```
MNDonut, line 37: you can set ids directly. re-read point 5 in the checklist

MNDonut (changes requested):

services/movie.py
```diff
@@ -0,0 +1,43 @@
+from django.db.models import QuerySet
+
+from db.models import Movie, Genre, Actor
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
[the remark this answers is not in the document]
HalynaRohatska: if I remove None or use | - the tests are not passed

MNDonut (changes requested):
Could you provide a screenshot of why tests fail? Code styling shouldn't impact the tests

services/cinema_hall.py
```diff
@@ -0,0 +1,19 @@
...
+def get_cinema_halls() -> QuerySet:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(
+        hall_name: str,
+        hall_rows: int,
+        hall_seats_in_row: int
+) -> CinemaHall:
+    return CinemaHall.objects.create(
```

oliasymak24 (changes requested):

services/movie.py
```diff
@@ -0,0 +1,42 @@
+from django.db.models import QuerySet
+from typing import Optional, List
+
+from db.models import Movie, Genre, Actor
```
MNDonut, line 4: different import groups should be separated by a blank line

services/movie.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def get_movies(
+        genres_ids: Optional[List[int]] = None,
+        actors_ids: Optional[List[int]] = None
```
MNDonut, line 9: let's use a modern approach of annotating unrequired params using | operator

services/movie.py
```diff
@@ -0,0 +1,42 @@
...
+        description=movie_description
+    )
+
+    if genres_ids:
+        Genre.objects.filter(id__in=genres_ids)
+        movie.genres.set(genres_ids)
+
+    if actors_ids:
+        Actor.objects.filter(id__in=actors_ids)
+        movie.actors.set(actors_ids)
```
MNDonut, line 42: why do you query Genre and Actor if you don't use them?

services/movie_session.py
```diff
@@ -0,0 +1,54 @@
+from _datetime import datetime
+from django.db.models import QuerySet
+from typing import Optional
+
+from db.models import MovieSession, CinemaHall, Movie
```
MNDonut, line 5: Fix imports
Also are you sure there is _datetime library?

services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+        movie_show_time: datetime,
+        movie_id: int,
+        cinema_hall_id: int
+) -> None:
+    movie_session = MovieSession(
+        show_time=movie_show_time,
+        cinema_hall_id=cinema_hall_id,
+        movie_id=movie_id
+    )
+    movie_session.save()
```
MNDonut, line 18: It's better to use MovieSession.objects.create

services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+        cinema_hall_id: Optional[int | None] = None
+) -> None:
+    session = MovieSession.objects.get(id=session_id)
+
+    if show_time:
+        session.show_time = show_time
+    if movie_id:
+        session.movie = Movie.objects.get(id=movie_id)
+    if cinema_hall_id:
+        session.cinema_hall = CinemaHall.objects.get(id=cinema_hall_id)
```
MNDonut, line 48: In order to not make 2 additional sql queries for movie and hall, you can set movie_id and cinema_hall_id directly. You can do it despite you haven't these fields in your model because Django is developed in the way that in the code you can use your added fields(movie, cinema_hall) and they contain related objects but under the hood in database ID is saved and column name is movie_id and cinema_hall_id because database doesn't know what is Django, related object but knows what is id

MNDonut (changes requested):

oliasymak24 (approved):

PR #907, opened 2024-08-26
Task Solution

achalaidiuk:


services/movie.py
```diff
@@ -0,0 +1,41 @@
...
+
+    if genres_ids:
+        movies = movies.filter(genres__id__in=genres_ids)
+    if actors_ids:
+        movies = movies.filter(actors__id__in=actors_ids)
+    return movies
+
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
```
Oleksl888, line 20: `pk` is preferable in django

services/movie.py
```diff
@@ -0,0 +1,41 @@
...
+        genres_ids: list | None = None
+) -> Movie:
+    created_movie = Movie.objects.create(
+        title=movie_title,
+        description=movie_description
+    )
+
+    if actors_ids:
+        actors = Actor.objects.filter(id__in=actors_ids)
+        created_movie.actors.set(actors)
```
Oleksl888, line 36: you can pass ids directly - no need to pass a queryset

services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+        movie_id=movie_id
+    )
+
+
+def get_movies_sessions(
+        session_date: datetime | None = None
+) -> QuerySet:
+
+    if session_date:
+        return MovieSession.objects.filter(show_time__date=session_date)
```
Oleksl888, line 25: this is good but the best solution would be with one `return`

services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+    return MovieSession.objects.all()
+
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
```
Oleksl888, line 30: pk

services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+    movie_session = get_movie_session_by_id(session_id)
+
+    if show_time is not None:
+        movie_session.show_time = show_time
+    if cinema_hall_id is not None:
+        movie_session.cinema_hall_id = cinema_hall_id
+    if movie_id is not None:
+        movie_session.movie_id = movie_id
+
+    movie_session.save()
```
Oleksl888, line 48: Can you refactor this function so that you will not need to get movie session if none of the update parameters has been passed?

services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+
+    return movie_session
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    MovieSession.objects.get(id=session_id).delete()
```
Oleksl888, line 54: Here you could call `get_movie_session_by_id` function to make code DRY

settings.py
```diff
@@ -21,6 +21,6 @@
 
 USE_I18N = True
 
-USE_TZ = False
+USE_TZ = True
```

Oleksl888 (changes requested):

services/cinema_hall.py
```diff
@@ -0,0 +1,19 @@
...
+def get_cinema_halls() -> QuerySet:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(
+        hall_name: str,
+        hall_rows: int,
+        hall_seats_in_row: int
+) -> CinemaHall:
+    return CinemaHall.objects.create(
```

oliasymak24 (changes requested):

PR #908, opened 2024-08-26
first_try

yashkunn:


services/movie.py
```diff
@@ -0,0 +1,42 @@
+from django.db.models import QuerySet, Q
+
+from db.models import Movie, Genre, Actor
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
MNDonut, line 8: list or none type

services/movie.py
```diff
@@ -0,0 +1,42 @@
...
+
+def get_movie_by_id(id_: int) -> Movie:
+    return Movie.objects.get(id=id_)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
MNDonut, line 32: list or none type

services/movie.py
```diff
@@ -0,0 +1,42 @@
...
+        return queryset
+    if genres_ids and actors_ids:
+        return queryset.filter(
+            Q(genres__id__in=genres_ids) & Q(actors__id__in=actors_ids)
+        ).distinct()
+    if genres_ids:
+        return queryset.filter(genres__id__in=genres_ids).distinct()
+    if actors_ids:
+        return queryset.filter(actors__id__in=actors_ids).distinct()
+    return queryset
```
MNDonut, line 21: you can simplify this code. isn't it easier to check if some parameter is passed and if yes, then filter queryset but don't return it and go to the next if statement?

MNDonut: also you can do distinct at the end in order do decrease sql queries to db

services/movie.py
```diff
@@ -0,0 +1,42 @@
...
+) -> Movie:
+    movie = Movie.objects.create(
+        title=movie_title,
+        description=movie_description
+    )
+    if genres_ids:
+        movie.genres.set(Genre.objects.filter(id__in=genres_ids))
+    if actors_ids:
+        movie.actors.set(Actor.objects.filter(id__in=actors_ids))
+    return movie
```
MNDonut, line 42: re-read the readme again. you can pass just ids without making redundant sql queries

services/movie_session.py
```diff
@@ -0,0 +1,61 @@
...
+) -> MovieSession:
+    try:
+        movie_session = MovieSession.objects.get(id=session_id)
+
+        if show_time:
+            movie_session.show_time = show_time
+        if movie_id is not None:
+            movie_session.movie_id = movie_id
+        if cinema_hall_id is not None:
+            movie_session.cinema_hall_id = cinema_hall_id
```
MNDonut, line 45: let's follow the same style and check `if show_time` everywhere instead of None check

services/movie_session.py
```diff
@@ -0,0 +1,61 @@
...
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: str = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
```
MNDonut, line 35: you know what to fix here ;)

services/movie_session.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def delete_movie_session_by_id(movie_session_id: int) -> None:
+    try:
+        movie_session = MovieSession.objects.get(id=movie_session_id)
+        movie_session.delete()
+    except MovieSession.DoesNotExist:
+        raise ValueError(
+            f"MovieSession with id {movie_session_id} does not exist."
+        )
```
MNDonut, line 61: 1. You have def get_movie_session_by_id to get movie session by id
2. Just a tip for the future: I recommend to use .filter() instead of get because get will raise an error if movie session doesn't exist and you need always add additional try blocks and you won't be able to delete this record while using filter() will not raise an error + will delete nothing if object doesn't exist

MNDonut (changes requested):

services/movie_session.py
```diff
@@ -52,10 +52,10 @@ def update_movie_session(
...
 def delete_movie_session_by_id(movie_session_id: int) -> None:
-    try:
-        movie_session = MovieSession.objects.get(id=movie_session_id)
+    movie_session = MovieSession.objects.filter(id=movie_session_id)
+    if movie_session.exists():
         movie_session.delete()
-    except MovieSession.DoesNotExist:
+    else:
         raise ValueError(
             f"MovieSession with id {movie_session_id} does not exist."
```
MNDonut, line 60: No need to check if record exists when you use filter(). Filter will return an empty queryset and delete will not delete anything. This is why we often use filter instead of get

services/movie.py
```diff
@@ -0,0 +1,36 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] | None = None,
+        actors_ids: list[int] | None = None
+) -> QuerySet | Movie:
```
MNDonut, line 9: queryset or movie? or queryset of movies?

services/movie_session.py
```diff
@@ -0,0 +1,61 @@
...
+        cinema_hall_id: int
+) -> MovieSession:
+    return MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> MovieSession:
```
MNDonut, line 18: str or None?

settings.py
```diff
@@ -21,6 +21,6 @@
 
 USE_I18N = True
 
-USE_TZ = False
+USE_TZ = True
```

MNDonut (changes requested):

services/movie.py
MNDonut: still not fixed

services/movie_session.py
MNDonut: still not fixed

MNDonut (changes requested):
Also tests fail

db/models.py
```diff
@@ -1,6 +1,39 @@
 from django.db import models
 
 
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField("Actor")
```
Oleksl888, line 7: related names are missing here

db/models.py
```diff
@@ -1,6 +1,39 @@
...
+        return self.name
+
+    @property
+    def capacity(self) -> int:
+        return self.rows * self.seats_in_row
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey("CinemaHall", on_delete=models.CASCADE)
```
Oleksl888, line 29: You don't need to use quotes on class name here + related names are missing

services/movie.py
```diff
@@ -0,0 +1,36 @@
...
+    queryset = Movie.objects.all()
+    if genres_ids:
+        queryset = queryset.filter(genres__id__in=genres_ids)
+    if actors_ids:
+        queryset = queryset.filter(actors__id__in=actors_ids)
+    return queryset.distinct()
+
+
+def get_movie_by_id(id_: int) -> Movie:
+    return Movie.objects.get(id=id_)
```
Oleksl888, line 19: Why not call this parameter `movie_id`?
Also `pk` is preferable way to query in django

services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+    return MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
+
+
+def get_movies_sessions(session_date: str | None = None) -> MovieSession:
+    if session_date:
+        date_object = datetime.strptime(session_date, "%Y-%m-%d").date()
```
Oleksl888, line 20: you can pass date without conversion here - django will take care of everything under the hood

services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+
+        return MovieSession.objects.filter(show_time__date=date_object)
```
Oleksl888, line 22: you can refactor this with only one return

services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+
+    return MovieSession.objects.all()
+
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
```
Oleksl888, line 28: pk

services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: str | None = None,
+        movie_id: int | None = None,
+        cinema_hall_id: int | None = None
+) -> MovieSession:
+    try:
+        movie_session = MovieSession.objects.get(id=session_id)
```
Oleksl888, line 38: Try to implement this:

1.     Create an empty dict
2. Check if any of the optional arguments are passed to this function. In case if there are - you will add them to the dict.
3. In the end check if your dict has some values to update and if yes you will call update on queryset, and if there are not any you don't do anything.

services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+            movie_session.cinema_hall_id = cinema_hall_id
+        movie_session.save()
+
+        return movie_session
+    except MovieSession.DoesNotExist:
+        raise ValueError(f"MovieSession with id {session_id} does not exist")
+
+
+def delete_movie_session_by_id(movie_session_id: int) -> None:
+    MovieSession.objects.filter(id=movie_session_id).delete()
```
Oleksl888, line 54: You can reuse get_movie_session_by_id function here to keep your code DRY?

settings.py
```diff
@@ -15,6 +15,7 @@
     }
 }
 
+
```
Oleksl888, line 18: please revert changes to this file

Oleksl888 (changes requested):

oliasymak24 (approved):

PR #909, opened 2024-08-27
implement py-movie-moviesession-cinemahall

ValeriiaKyr:


services/movie.py
```diff
@@ -1,11 +1,13 @@
...
 from django.db.models import QuerySet
 
 from db.models import Movie
 
 
 def get_movies(
-        genres_ids: list[int] = None,
-        actors_ids: list[int] = None,
+        genres_ids: Union[List[int], None] = None,
+        actors_ids: Union[List[int], None] = None,
```
MNDonut, line 10: just list | None is enough

MNDonut (changes requested):

MNDonut (approved):

PR #910, opened 2024-08-27
Solution

olena4reunited:


services/movie.py
```diff
@@ -0,0 +1,45 @@
...
+        genres_ids: list | None = None,
+        actors_ids: list | None = None
+) -> QuerySet:
+    movies = Movie.objects.all()
+
+    if genres_ids is not None:
+        movies = movies.filter(genres__id__in=genres_ids).distinct()
+
+    if actors_ids is not None:
+        movies = movies.filter(actors__id__in=actors_ids).distinct()
```
MNDonut, line 17: just `if actors_ids` is enough because python checks if it isn't None, False, empty string or list, etc

services/movie.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def get_movies(
+        genres_ids: list | None = None,
+        actors_ids: list | None = None
+) -> QuerySet:
+    movies = Movie.objects.all()
+
+    if genres_ids is not None:
+        movies = movies.filter(genres__id__in=genres_ids).distinct()
```
MNDonut, line 14: You can do distinct at the end to not overwhelm sql queries

services/movie.py
```diff
@@ -0,0 +1,45 @@
...
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: Optional[list] = None,
+        actors_ids: Optional[list] = None
```
MNDonut, line 30: let's follow the same code styling and use | operator

services/movie.py
```diff
@@ -0,0 +1,45 @@
...
+        description=movie_description,
+    )
+
+    if genres_ids is not None:
+        for genre_id in genres_ids:
+            movie.genres.add(genre_id)
+
+    if actors_ids is not None:
+        for actor_id in actors_ids:
+            movie.actors.add(actor_id)
```
MNDonut, line 43: 1. just `if actors_ids` is enough
2. also you can use .set() method that works like your loops(sets all ids at the same time)

services/movie_session.py
```diff
@@ -0,0 +1,60 @@
+from datetime import datetime
+from django.db.models import QuerySet
+from typing import Optional
```
MNDonut, line 3: different import groups should be separated by a b;aml ;ome

services/movie_session.py
```diff
@@ -0,0 +1,60 @@
...
+) -> MovieSession:
+    return MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id,
+    )
+
+
+def get_movies_sessions(
+        session_date: Optional[str] = None
```
MNDonut, line 21: let's follow the same code styling and use | operator

services/movie_session.py
```diff
@@ -0,0 +1,60 @@
...
+) -> QuerySet:
+    if session_date:
+        return MovieSession.objects.filter(
+            show_time__date=datetime.strptime(session_date, "%Y-%m-%d")
+        )
+    else:
```
MNDonut, line 27: no need for else because if statement has return

services/movie_session.py
```diff
@@ -0,0 +1,60 @@
...
+) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: Optional[datetime] = None,
+        movie_id: Optional[int] = None,
+        cinema_hall_id: Optional[int] = None
+) -> MovieSession:
```
MNDonut, line 42: you know what to do here ;)

services/movie_session.py
```diff
@@ -0,0 +1,60 @@
...
+    session_to_update = MovieSession.objects.get(pk=session_id)
+
+    if show_time is not None:
+        session_to_update.show_time = show_time
+    if movie_id is not None:
+        session_to_update.movie_id = movie_id
+    if cinema_hall_id is not None:
+        session_to_update.cinema_hall_id = cinema_hall_id
```
MNDonut, line 50: just `if <variable> is enough`

services/movie_session.py
```diff
@@ -0,0 +1,60 @@
...
+
+    session_to_update.save()
+
+    return session_to_update
+
+
+def delete_movie_session_by_id(
+        session_id: int
+) -> None:
+    MovieSession.objects.get(id=session_id).delete()
```
MNDonut, line 60: 1. You have def get_movie_session_by_id to get movie session by id
2. Personally I recommend to use .filter() instead of get because get will raise an error if movie session doesn't exist and you won't be able to delete this record while using filter() will not raise an error + will delete nothing if object doesn't exist :)

MNDonut (changes requested):

MNDonut (approved):
Perfect!

PR #911, opened 2024-08-27
write movie-moviesession-cinemahall

vladislav-tsybuliak1:


services/movie.py
```diff
@@ -0,0 +1,47 @@
+from typing import Optional
+
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: Optional[list[int]] = None,
+        actors_ids: Optional[list[int]] = None
```
MNDonut, line 10: let's use | opeator instead of Optional. fix everywhere

services/movie.py
```diff
@@ -0,0 +1,47 @@
...
+        )
+    if genres_ids:
+        return query_set.filter(
+            genres__id__in=genres_ids
+        )
+    if actors_ids:
+        return query_set.filter(
+            actors__id__in=actors_ids
+        )
+    return query_set
```
MNDonut, line 26: let's filter queryset instead of making return. in this case each if statement overrides the initial queryset with some filter and then you don't need the first condition

MNDonut (changes requested):

services/movie.py
```diff
@@ -1,26 +1,24 @@
...
+        query_set = query_set.filter(
             genres__id__in=genres_ids
         )
-    if actors_ids:
-        return query_set.filter(
+    elif actors_ids:
+        query_set = query_set.filter(
             actors__id__in=actors_ids
         )
     return query_set
```
MNDonut: you didn't get it.
Let's remove the first if and leave the last to as "if", not "elif". In this case if both parameters are passed, python will get inside each if statement and appropriate filter will be added

MNDonut (changes requested):

services/movie.py
vladislav-tsybuliak1: I got it after I'd commited. Now everything should be ok

MNDonut (approved):
The fastest fixes on the wild west!!

PR #913, opened 2024-08-27
Solution Politov

Alexandr-Politov:


db/models.py
```diff
@@ -14,3 +14,39 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
```
Oleksl888, line 22: what about related names?

db/models.py
```diff
@@ -14,3 +14,39 @@ class Actor(models.Model):
...
+
+    def __str__(self) -> str:
+        return self.name
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall,
+                                    on_delete=models.CASCADE,
+                                    related_name="movies")
```
Oleksl888, line 46: related name must represent a link to moviesession on another model

services/movie.py
```diff
@@ -0,0 +1,29 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(genres_ids: list[int] = None,
```
Oleksl888, line 6: annotate that these fields are Optional

services/movie.py
```diff
@@ -0,0 +1,29 @@
...
+    result = Movie.objects.all()
+    if genres_ids:
+        result = result.filter(genres__id__in=genres_ids)
+    if actors_ids:
+        result = result.filter(actors__id__in=actors_ids)
+    return result
+
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
```
Oleksl888, line 17: pk is preferable over id in django

services/movie.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def create_movie(movie_title: str,
+                 movie_description: str,
+                 genres_ids: list[int] = None,
```
Oleksl888, line 22: add optional annotation

services/movie_session.py
```diff
@@ -0,0 +1,43 @@
...
+
+from db.models import MovieSession, Movie, CinemaHall
+
+
+def create_movie_session(movie_show_time: str,
+                         movie_id: int,
+                         cinema_hall_id: int) -> None:
+    movie = Movie.objects.get(id=movie_id)
+    hall = CinemaHall.objects.get(id=cinema_hall_id)
+    MovieSession.objects.create(show_time=movie_show_time,
```
Oleksl888, line 11: Or you could do it in one line passing ids on movie session create `movie_id=movie_id`

services/movie_session.py
```diff
@@ -0,0 +1,43 @@
...
+                                cinema_hall=hall,
+                                movie=movie)
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
```
Oleksl888, line 16: add optional annotation everywhere please

services/movie_session.py
```diff
@@ -0,0 +1,43 @@
...
+    result = MovieSession.objects.all()
+    if session_date:
+        result = result.filter(show_time__date=session_date)
+    return result
+
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
```
Oleksl888, line 24: pk

services/movie_session.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def update_movie_session(session_id: int,
+                         show_time: str = None,
+                         movie_id: int = None,
+                         cinema_hall_id: int = None) -> None:
+    session = MovieSession.objects.get(id=session_id)
```
Oleksl888, line 31: Can you refactor this function so that you will not need to get movie session if none of the update parameters has been passed?

services/movie_session.py
```diff
@@ -0,0 +1,43 @@
...
+        session.show_time = show_time
+    if movie_id:
+        session.movie = Movie.objects.get(id=movie_id)
+    if cinema_hall_id:
+        session.cinema_hall = CinemaHall.objects.get(id=cinema_hall_id)
+    session.save()
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    session = MovieSession.objects.get(id=session_id)
```
Oleksl888, line 42: Here you could call get_movie_session_by_id function to make code DRY

settings.py
```diff
@@ -21,6 +21,6 @@
 
 USE_I18N = True
 
-USE_TZ = False
+USE_TZ = True
```

Oleksl888 (changes requested):

db/models.py
```diff
@@ -43,7 +43,7 @@ class MovieSession(models.Model):
     show_time = models.DateTimeField()
     cinema_hall = models.ForeignKey(CinemaHall,
                                     on_delete=models.CASCADE,
-                                    related_name="movies")
+                                    related_name="sessions")
     movie = models.ForeignKey(Movie,
                               on_delete=models.CASCADE,
                               related_name="sessions")
```
MNDonut, line 49: We don't write code in this way. Each parameter should be added on the next line

services/movie.py
```diff
@@ -1,10 +1,12 @@
...
+from typing import Optional
+
 
 from db.models import Movie
 
 
-def get_movies(genres_ids: list[int] = None,
-               actors_ids: list[int] = None) -> QuerySet:
+def get_movies(genres_ids: Optional[list[int]] = None,
+               actors_ids: Optional[list[int]] = None) -> QuerySet:
```
MNDonut, line 9: It's better to use | to decrease number of imports

services/movie.py
```diff
@@ -1,10 +1,12 @@
 from django.db.models import QuerySet
+from typing import Optional
+
 
 from db.models import Movie
```
MNDonut, line 5: there is a specific order of different imports + import groups should be splited by 1 blank line

services/movie_session.py
MNDonut: still not fixed @Alexandr-Politov

services/movie.py
MNDonut: Fix everywhere

db/models.py
MNDonut: Fix everywhere

services/movie_session.py
```diff
@@ -1,43 +1,46 @@
...
+    if any([show_time, movie_id, cinema_hall_id]):
+        session = MovieSession.objects.get(pk=session_id)
     if show_time:
         session.show_time = show_time
     if movie_id:
-        session.movie = Movie.objects.get(id=movie_id)
+        session.movie = Movie.objects.get(pk=movie_id)
     if cinema_hall_id:
-        session.cinema_hall = CinemaHall.objects.get(id=cinema_hall_id)
+        session.cinema_hall = CinemaHall.objects.get(pk=cinema_hall_id)
```
MNDonut, line 40: you can avoid making additional 3 sql queries setting session.**movie_id** for example

MNDonut (changes requested):

services/movie_session.py
```diff
@@ -1,21 +1,24 @@
...
     movie = Movie.objects.get(pk=movie_id)
     hall = CinemaHall.objects.get(pk=cinema_hall_id)
-    MovieSession.objects.create(show_time=movie_show_time,
-                                cinema_hall=hall,
-                                movie=movie)
+    MovieSession.objects.create(
+        show_time=movie_show_time,
+        cinema_hall=hall,
+        movie=movie
+    )
```
MNDonut, line 18: Still not fixed. In order to not make 2 additional sql queries for movie and hall, you can set movie_id and cinema_hall_id directly. You can do it despite you haven't these fields in your model because Django is developed in the way that in the code you can use your added fields(movie, cinema_hall) and they contain related objects but under the hood in database ID is saved and column name is movie_id and cinema_hall_id because database doesn't know what is Django, for is related object but knows what is id

MNDonut (changes requested):

services/movie_session.py
```diff
@@ -0,0 +1,49 @@
...
+        session.show_time = show_time
+    if movie_id:
+        session.movie_id = movie_id
+    if cinema_hall_id:
+        session.cinema_hall_id = cinema_hall_id
+    session.save()
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    session = get_movie_session_by_id(session_id)
```
oliasymak24, line 48: Np need to add this additional variable

services/movie_session.py
```diff
@@ -0,0 +1,49 @@
...
+) -> None:
+    if any([show_time, movie_id, cinema_hall_id]):
+        session = MovieSession.objects.get(pk=session_id)
+    if show_time:
+        session.show_time = show_time
+    if movie_id:
+        session.movie_id = movie_id
+    if cinema_hall_id:
+        session.cinema_hall_id = cinema_hall_id
+    session.save()
```
oliasymak24, line 44: Try to use update function here

oliasymak24 (changes requested):

MNDonut (approved):

PR #919, opened 2024-08-28
Solution

mihavryliuk:


services/movie.py
```diff
@@ -0,0 +1,39 @@
...
+        queryset = queryset.filter(genres__id__in=genres_ids)
+
+    if actors_ids:
+        queryset = queryset.filter(actors__id__in=actors_ids)
+
+    return queryset.distinct()
+
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
```
Oleksl888, line 20: using pk is preferable in django

services/movie_session.py
```diff
@@ -0,0 +1,52 @@
...
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
+    sessions = MovieSession.objects.all()
+    if session_date:
+        sessions = sessions.filter(
+            show_time__date=datetime.strptime(session_date, "%Y-%m-%d")
```
Oleksl888, line 24: you can pass date as is - django will take care of the formatting and conversion under the hood

services/movie_session.py
```diff
@@ -0,0 +1,52 @@
...
+        movie_session.movie_id = movie_id
+    if cinema_hall_id:
+        movie_session.cinema_hall_id = cinema_hall_id
+    movie_session.save()
+
+    return movie_session
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    return MovieSession.objects.get(id=session_id).delete()
```
Oleksl888, line 52: Here you could reuse `get_movie_session_by_id` function to make code DRY
Also no need to return result from deletion

Oleksl888 (changes requested):

services/movie_session.py
```diff
@@ -0,0 +1,52 @@
...
+    MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
+    sessions = MovieSession.objects.all()
+    if session_date:
```
Oleksl888, line 23: why like this?

services/movie_session.py
```diff
@@ -0,0 +1,52 @@
...
+
+def update_movie_session(
+        session_id: int,
+        show_time: Optional[datetime] = None,
+        movie_id: Optional[int] = None,
+        cinema_hall_id: Optional[int] = None
+) -> None:
+    if not any([show_time, movie_id, cinema_hall_id]):
+        return
+    movie_session = get_movie_session_by_id(session_id)
```
Oleksl888, line 41: what if you create a dict, and then you will check if any of the optional arguments are passed to this function. In case if there are - you will add them to the dict. Then after you checked if your dict has some values to update you will call update on queryset, and if there are not any you don't do anything

Oleksl888 (changes requested):

services/cinema_hall.py
```diff
@@ -0,0 +1,19 @@
...
+def get_cinema_halls() -> QuerySet:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(
+        hall_name: str,
+        hall_rows: int,
+        hall_seats_in_row: int
+) -> CinemaHall:
+    return CinemaHall.objects.create(
```

oliasymak24 (changes requested):

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
...
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
+
+
+def get_movies_sessions(session_date: Optional[str] = None) -> QuerySet:
+    sessions = MovieSession.objects.all()
+    if session_date:
+        if session_date:
```
Oleksl888, line 24: why have two nested if condition here? please remove

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
...
+
+    if show_time:
+        update_fields["show_time"] = show_time
+    if movie_id:
+        update_fields["movie_id"] = movie_id
+    if cinema_hall_id:
+        update_fields["cinema_hall_id"] = cinema_hall_id
+
+    if update_fields:
+        MovieSession.objects.filter(pk=session_id).update(**update_fields)
```
Oleksl888, line 49: bravo :clap:

Oleksl888 (approved):
Well done! But please fix movie_session.py -> line 23

PR #921, opened 2024-08-28
Solution

dpiuro:


settings.py
```diff
@@ -21,6 +21,6 @@
 
 USE_I18N = True
 
-USE_TZ = False
```

services/movie_session.py
```diff
@@ -21,22 +21,27 @@ def create_movie_session(
 def get_movies_sessions(
         session_date: Optional[datetime.date] = None
```
MNDonut, line 22: let's use | operator everywhere as it's a modern approach

MNDonut (changes requested):

services/cinema_hall.py
```diff
@@ -0,0 +1,19 @@
...
+def get_cinema_halls() -> QuerySet[CinemaHall]:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(
+    hall_name: str,
+    hall_rows: int,
+    hall_seats_in_row: int
+) -> CinemaHall:
+    return CinemaHall.objects.create(
```

services/movie_session.py
```diff
@@ -0,0 +1,59 @@
...
+    session = MovieSession.objects.get(pk=session_id)
+
+    if show_time:
+        session.show_time = show_time
+    if movie_id:
+        session.movie_id = movie_id
+    if cinema_hall_id:
+        session.cinema_hall_id = cinema_hall_id
+
+    session.save()
```
oliasymak24, line 52: Try to use update function here)

services/movie_session.py
```diff
@@ -0,0 +1,59 @@
...
+
+    return session
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    session = get_movie_session_by_id(session_id)
```
oliasymak24, line 58: No need to create separate variable

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #923, opened 2024-08-28
'Solution'

ABaidin:


ABaidin:
oh, and why did i think there were no tests... 
I'm gonna fix everything now

settings.py
oliasymak24, line 1: Don't need to commit this file

services/movie_session.py
```diff
@@ -0,0 +1,55 @@
...
+
+from db.models import MovieSession, Movie, CinemaHall
+
+
+def create_movie_session(
+        movie_show_time: datetime,
+        movie_id: int,
+        cinema_hall_id: int
+) -> MovieSession:
+    return MovieSession.objects.create(
```
oliasymak24, line 12: You don’t need to `return` anything in create/update/delete methods, check everywhere

services/movie.py
```diff
@@ -0,0 +1,41 @@
+from typing import List
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: List[int] = None,
+        actors_ids: List[int] = None
+) -> List[Movie]:
```
oliasymak24, line 9: This i a wrong annotation, fix everywhere

oliasymak24 (changes requested):

services/cinema_hall.py
```diff
@@ -0,0 +1,19 @@
+from typing import List
+
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> List[CinemaHall]:
```
MNDonut, line 6: it's queryset, not list

services/movie.py
MNDonut: @ABaidin still not fixed
Also genres_ids and actors_ids are lists or None

services/movie.py
```diff
@@ -0,0 +1,39 @@
...
+        actors_ids: List[int] = None
+) -> List[Movie]:
+    queryset = Movie.objects.all()
+
+    if genres_ids:
+        queryset = queryset.filter(genres__id__in=genres_ids).distinct()
+    if actors_ids:
+        queryset = queryset.filter(actors__id__in=actors_ids).distinct()
+
+    return queryset
```
MNDonut, line 17: let's call distinct at the end

services/movie.py
```diff
@@ -0,0 +1,39 @@
...
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: List[int] = None,
+        actors_ids: List[int] = None
```
MNDonut, line 28: list or None

services/movie_session.py
```diff
@@ -0,0 +1,55 @@
...
+        movie_session_id: int,
+) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: datetime = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
```
MNDonut, line 41: datetime/int or None

services/movie_session.py
```diff
@@ -0,0 +1,55 @@
...
+) -> None:
+    session = MovieSession.objects.get(id=session_id)
+    if show_time:
+        session.show_time = show_time
+    if movie_id:
+        session.movie = Movie.objects.get(id=movie_id)
+    if cinema_hall_id:
+        session.cinema_hall = CinemaHall.objects.get(id=cinema_hall_id)
```
MNDonut, line 49: let's use ids directly(re-read checklist again). you can set show_time_id and movie_id fields with ids you get so that you don't need to make additional queries to database

settings.py
```diff
@@ -21,6 +21,6 @@
 
 USE_I18N = True
 
-USE_TZ = False
+USE_TZ = True
```

MNDonut (changes requested):

MNDonut (approved):

PR #924, opened 2024-08-28
first

egor250330:


db/models.py
```diff
@@ -14,3 +14,38 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
```
Oleksl888, line 22: related names missing

db/models.py
```diff
@@ -14,3 +14,38 @@ class Actor(models.Model):
...
+    def capacity(self) -> int:
+        return self.rows * self.seats_in_row
+
+    def __str__(self) -> str:
+        return self.name
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall, on_delete=models.CASCADE)
```
Oleksl888, line 44: related names missing

services/movie.py
```diff
@@ -0,0 +1,52 @@
...
+
+    if actors_ids:
+        movies = movies.filter(actors__in=actors_ids)
+
+    return movies
+
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    try:
+        movie = Movie.objects.get(id=movie_id)
```
Oleksl888, line 27: `pk` is preferable key to use in django

services/movie.py
```diff
@@ -0,0 +1,52 @@
...
+        return movie
+    except ObjectDoesNotExist:
```
Oleksl888, line 29: This is not so good pattern - if record doesn't exist the warning should happen, if you silently return None instead of a Movie it may lead to undepredictable program behaviour later

services/movie.py
```diff
@@ -0,0 +1,52 @@
...
+        movie_title: str,
+        movie_description: str,
+        genres_ids: list | None = None,
+        actors_ids: list | None = None
+) -> Movie:
+    movie = Movie.objects.create(
+        title=movie_title,
+        description=movie_description
+    )
+    if genres_ids is not None:
```
Oleksl888, line 43: No need to compare to None explicitly

services/movie.py
```diff
@@ -0,0 +1,52 @@
...
+        genres = Genre.objects.filter(id__in=genres_ids)
+        movie.genres.set(genres)
```
Oleksl888, line 45: you can pass id's here no need to pass queryset

services/movie.py
```diff
@@ -0,0 +1,52 @@
...
+
+    if actors_ids is not None:
+        actors = Actor.objects.filter(id__in=actors_ids)
+        movie.actors.set(actors)
+
+    movie.save()
```
Oleksl888, line 51: no need to call .save() since we are using manager

services/movie_session.py
```diff
@@ -0,0 +1,57 @@
...
+def get_movies_sessions(session_date: str | None = None) -> QuerySet:
+    new_set = MovieSession.objects.all()
+    if session_date:
+        new_set = new_set.filter(show_time__date=session_date)
+
+    return new_set
+
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
```
Oleksl888, line 32: pk

services/movie_session.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: datetime = None,
```
Oleksl888, line 37: add annotation to show these params are optional

services/movie_session.py
```diff
@@ -0,0 +1,57 @@
...
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> MovieSession:
+    movie_session = MovieSession.objects.get(id=session_id)
```
Oleksl888, line 41: Can you refactor this function so that you will not need to get movie session if none of the update parameters has been passed?

services/movie_session.py
```diff
@@ -0,0 +1,57 @@
...
+
+    if cinema_hall_id:
+        movie_session.cinema_hall_id = cinema_hall_id
+
+    movie_session.save()
+    return movie_session
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    MovieSession.objects.filter(id=session_id).delete()
```
Oleksl888, line 57: Here you could reuse `get_movie_session_by_id` function to make code DRY

Oleksl888 (changes requested):

services/movie_session.py
```diff
@@ -0,0 +1,66 @@
...
+def create_movie_session(
+        movie_show_time: datetime,
+        movie_id: int,
+        cinema_hall_id: int
+) -> MovieSession:
+    movie = Movie.objects.get(pk=movie_id)
+    cinema_hall = CinemaHall.objects.get(pk=cinema_hall_id)
+    return MovieSession.objects.create(
+        movie=movie,
+        cinema_hall=cinema_hall,
```
Oleksl888, line 18: here you can pass ids directly instead of creating objects first - use cinema_hall_id

services/movie_session.py
```diff
@@ -0,0 +1,66 @@
...
+def get_movies_sessions(session_date: Optional[str] = None) -> QuerySet:
+    queryset = MovieSession.objects.all()
+    if session_date:
+        queryset = queryset.filter(show_time__date=session_date)
+    return queryset
+
+
+def get_movie_session_by_id(movie_session_id: int) -> Optional[MovieSession]:
+    try:
+        return MovieSession.objects.get(pk=movie_session_id)
```
Oleksl888, line 32: This is not so good pattern - if record doesn't exist, warning should be raised and execution must be stopped, if you return None instead of a MovieSession it may lead to unpredictable behavior in other parts of the program

services/movie_session.py
```diff
@@ -0,0 +1,66 @@
...
+        return None
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: Optional[datetime] = None,
+        movie_id: Optional[int] = None,
+        cinema_hall_id: Optional[int] = None
+) -> Optional[MovieSession]:
+    if not any([show_time, movie_id, cinema_hall_id]):
```
Oleksl888, line 43: Create an empty dict. Then you will check if any of the optional arguments are passed to this function. In case if there are - you will add them to the dict. 
Then after all that - make a condition that if your dict has some values to update you will call update on queryset, and if there are not any you don't do anything

services/movie_session.py
```diff
@@ -0,0 +1,66 @@
...
+
+    if cinema_hall_id:
+        movie_session.cinema_hall = CinemaHall.objects.get(pk=cinema_hall_id)
+
+    movie_session.save()
+    return movie_session
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    movie_session = get_movie_session_by_id(session_id)
```
Oleksl888, line 64: just do this `get_movie_session_by_id(session_id).delete()`

Oleksl888 (changes requested):

services/movie_session.py
```diff
@@ -0,0 +1,63 @@
...
+        cinema_hall_id: Optional[int] = None
+) -> Optional[MovieSession]:
+    if not any([show_time, movie_id, cinema_hall_id]):
+        return None
+
+    movie_session = get_movie_session_by_id(session_id)
+    if movie_session is None:
+        return None
+
+    update_fields = {}
```
Oleksl888, line 45: You can just leave lines 45-57. No need for any extra checks, no need to return queryset in this function

Oleksl888 (changes requested):

MNDonut (approved):

PR #925, opened 2024-08-28
Solution

Ivan-Shakhman:
Present my first Solution of this task


db/models.py
```diff
@@ -14,3 +14,39 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
```
Oleksl888, line 22: related names are missing

db/models.py
```diff
@@ -14,3 +14,39 @@ class Actor(models.Model):
...
+        return self.seats_in_row * self.rows
+
+    def __str__(self) -> str:
+        return self.name
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(
+        CinemaHall, on_delete=models.SET_NULL, related_name="halls", null=True
```
Oleksl888, line 45: would make sense to use CASCADE here since movie session cannot happen without a movie or without location

db/models.py
```diff
@@ -14,3 +14,39 @@ class Actor(models.Model):
...
+    )
+    movie = models.ForeignKey(
```
Oleksl888, line 47: Also check the related names on you Fkeys - this name should present a link to `MovieSessions` on a different object

services/movie.py
```diff
@@ -0,0 +1,37 @@
+from django.db.models import QuerySet
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list = None,
```
Oleksl888, line 6: Annotate optional parameters - check in other functions too

services/movie.py
```diff
@@ -0,0 +1,37 @@
+from django.db.models import QuerySet
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list = None,
+        actors_ids: list = None
+) -> QuerySet:
+    if genres_ids and actors_ids:
```
Oleksl888, line 9: refactor this so you use only two if statements and one return

services/movie.py
```diff
@@ -0,0 +1,37 @@
...
+        )
+    if genres_ids:
+        return Movie.objects.filter(genres__in=genres_ids).distinct()
+    if actors_ids:
+        return Movie.objects.filter(actors__in=actors_ids).distinct()
+    return Movie.objects.all()
+
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
```
Oleksl888, line 21: using pk is preferable in django

services/movie.py
```diff
@@ -0,0 +1,37 @@
...
+        movie_title: str,
+        movie_description: str,
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
+) -> Movie:
+    created_movie = Movie.objects.create(
+        title=movie_title, description=movie_description
+    )
+    if genres_ids:
+        created_movie.genres.add(*genres_ids)
```
Oleksl888, line 34: refer to the checklist for proper handling m2m fields

services/movie_session.py
```diff
@@ -0,0 +1,58 @@
...
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id,
+    )
+
+
+def get_movies_sessions(session_date: str = "") -> MovieSession | None:
+    if session_date:
+        year, month, day = session_date.split("-")
+        return MovieSession.objects.filter(
```
Oleksl888, line 21: refactor this to use only one return. 
Also you may pass string as is without conversion - django will take care of everything under the hood

services/movie_session.py
```diff
@@ -0,0 +1,58 @@
...
+            show_time__date=date(int(year), int(month), int(day)),
+        )
+    return MovieSession.objects.all()
+
+
+def get_movie_session_by_id(movie_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_id)
```
Oleksl888, line 28: pk

services/movie_session.py
```diff
@@ -0,0 +1,58 @@
...
+def update_movie_session(
+        session_id: int,
+        show_time: datetime = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> None:
+    try:
+        movie_session = MovieSession.objects.get(id=session_id)
+    except ObjectDoesNotExist:
+        print(f"Movie Session with ID {session_id} does not exist")
```
Oleksl888, line 40: This is not so good pattern - if record doesn't exist, warning should be raised and execution must be stopped, if you return None instead of a Movie it may lead to unpredictable behavior in other parts of the program

services/movie_session.py
```diff
@@ -0,0 +1,58 @@
...
+    else:
+        if show_time:
+            movie_session.show_time = show_time
+        if cinema_hall_id:
```
Oleksl888, line 44: Can you refactor this so that you will not need to get movie session if none of the update parameters has been passed?

services/movie_session.py
```diff
@@ -0,0 +1,58 @@
...
+            )
+        if movie_id:
+            movie_session.movie = Movie.objects.get(id=movie_id)
+        movie_session.save()
+        return movie_session
+    return None
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    movie_session = MovieSession.objects.get(id=session_id)
```
Oleksl888, line 56: Here you could reuse `get_movie_session_by_id` function to make code DRY.
Also result of deletion is not None so either don't return anything or change annotation for return type

Oleksl888 (changes requested):

db/models.py
```diff
@@ -14,3 +14,40 @@ class Actor(models.Model):
...
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(
+        CinemaHall, on_delete=models.CASCADE, related_name="movie_sessions"
+    )
+    movie = models.ForeignKey(
+        Movie, on_delete=models.SET_NULL,
+        related_name="movie_sessions", null=True
```
MNDonut, line 49: we don't pass multiple parameters in this way. Either all in 1 line or 1 parameter - 1 line

services/movie.py
MNDonut: @Ivan-Shakhman still not fixed

services/movie.py
```diff
@@ -0,0 +1,37 @@
+from typing import Optional
+
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: Optional[list[int]] = None,
+        actors_ids: Optional[list[int]] = None
```
MNDonut, line 10: we recommend to use | operator because it's a modern way from python 3.10

services/movie.py
```diff
@@ -0,0 +1,37 @@
...
+
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: Optional[list[int]] = None,
+        actors_ids: Optional[list[int]] = None
+) -> QuerySet:
```
MNDonut, line 11: specify what kind of queryset this function returns

services/movie.py
```diff
@@ -0,0 +1,37 @@
...
+    movies = Movie.objects.all()
+    if genres_ids:
+        movies = movies.filter(genres__in=genres_ids)
+    if actors_ids:
+        movies = movies.filter(actors__in=actors_ids)
+    return movies.distinct()
```
MNDonut, line 17: 1 blank line after the last if

services/movie_session.py
MNDonut: @Ivan-Shakhman still not fixed

services/movie_session.py
```diff
@@ -0,0 +1,61 @@
...
+    else:
+        if show_time:
+            movie_session.show_time = show_time
+        if cinema_hall_id:
+            movie_session.cinema_hall = (
+                CinemaHall.objects.get(id=cinema_hall_id)
+            )
+        if movie_id:
+            movie_session.movie = Movie.objects.get(id=movie_id)
+        movie_session.save()
```
MNDonut, line 53: let's pass id of cinema_hall and movie instead of records(use cinema_hall_id and movie_id fields)

services/movie_session.py
MNDonut: @Ivan-Shakhman still not fixed

MNDonut (changes requested):

services/movie_session.py
```diff
@@ -19,43 +19,44 @@ def create_movie_session(
 
 
 def get_movies_sessions(session_date: str = "") -> MovieSession | None:
```
MNDonut, line 21: are you sure it returns movie session object or None?

services/movie_session.py
```diff
@@ -19,43 +19,44 @@ def create_movie_session(
...
 def get_movie_session_by_id(movie_id: int) -> MovieSession:
-    return MovieSession.objects.get(id=movie_id)
+    return MovieSession.objects.get(pk=movie_id)
 
 
 def update_movie_session(
         session_id: int,
         show_time: Optional[datetime] = None,
         movie_id: Optional[int] = None,
         cinema_hall_id: Optional[int] = None
```
MNDonut, line 38: still not fixed

services/movie_session.py
```diff
@@ -19,43 +19,44 @@ def create_movie_session(
...
             movie_session.show_time = show_time
         if cinema_hall_id:
             movie_session.cinema_hall = (
-                CinemaHall.objects.get(id=cinema_hall_id)
+                CinemaHall.objects.get(pk=cinema_hall_id)
             )
         if movie_id:
-            movie_session.movie = Movie.objects.get(id=movie_id)
+            movie_session.movie = Movie.objects.get(pk=movie_id)
         movie_session.save()
```
MNDonut, line 57: Re-read my last comment about it. The issue isn't in pk, it's because you make redundant queries. Use the fields I mentioned earlier

services/movie_session.py
```diff
@@ -19,43 +19,44 @@ def create_movie_session(
...
-    return None
+        return movie_session.cinema_hall_id, movie_session.movie_id
 
 
-def delete_movie_session_by_id(session_id: int) -> None:
-    movie_session = MovieSession.objects.get(id=session_id)
-    movie_session.delete()
-    return movie_session
+def delete_movie_session_by_id(session_id: int) -> MovieSession:
+    return MovieSession.objects.get(pk=session_id).delete()
```
MNDonut, line 62: follow DRY and use get_movie_session_by_id

db/models.py
```diff
@@ -14,3 +14,40 @@ class Actor(models.Model):
...
```
Oleksl888, line 49: So you would delete this record if related cinema hall is deleted, and you will keep the record in case if related movie is deleted. Can you please tell why did you make this choice?

services/movie.py
Oleksl888: using set() method is preferable here

services/movie_session.py
Oleksl888: Try a different approach:
1. create a dictionary, 
2.  check if any of the optional arguments are passed to this function. In case if there are - you will add them to the dict. 
3. In the end check if your dict has some values to update and if yes you will call update on queryset, and if there are not any you don't do anything

Oleksl888 (changes requested):

services/movie_session.py
```diff
@@ -0,0 +1,64 @@
...
+        cinema_hall_id: int,
+) -> MovieSession:
+    return MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id,
+    )
+
+
+def get_movies_sessions(session_date: str = "") -> QuerySet:
```
Oleksl888, line 22: here missing annotation that session_date is optional

Oleksl888, line 22: here missing annotation that session_date is optional

services/movie_session.py
```diff
@@ -0,0 +1,64 @@
...
+def update_movie_session(
+        session_id: int,
+        show_time: datetime | None = None,
+        movie_id: int | None = None,
+        cinema_hall_id: int | None = None
+) -> tuple[Any, Any]:
+    update_data = {}
+    if show_time:
+        update_data["show_time"] = show_time
+    if cinema_hall_id:
```
Oleksl888, line 44: You don't have return anything in this method

services/movie_session.py
```diff
@@ -0,0 +1,64 @@
...
+        update_data["cinema_hall_id"] = cinema_hall_id
+    if movie_id:
+        update_data["movie_id"] = movie_id
+
+    if update_data:
+        try:
```
Oleksl888, line 50: just do 
```
if update_data:
        MovieSession.objects.filter(pk=session_id).update(**update_data)

Comment (Oleksl888) [reply services/movie_session.py:50]: you don't need to try except here - if movie session doesn't exist with this pk - just let the error happen (in this scenario is acceptable)

**Review by Oleksl888 — CHANGES_REQUESTED** [review]

File: services/cinema_hall.py [hunk 26]
```diff
@@ -0,0 +1,19 @@
...
+def get_cinema_halls() -> QuerySet:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(
+        hall_name: str,
+        hall_rows: int,
+        hall_seats_in_row: int
+) -> CinemaHall:
+    return CinemaHall.objects.create(
```
oliasymak24, line 15: You don't need any return statement in create/udate/delete functions, check everywhere

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
...
+        cinema_hall_id: int,
+) -> MovieSession:
+    return MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id,
+    )
+
+
+def get_movies_sessions(session_date: str = "") -> QuerySet:
```
oliasymak24, line 20: session_date must be optional

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #926, opened 2024-08-28
Resolve

ihorutkin:


db/models.py
```diff
@@ -14,3 +14,43 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
```
Oleksl888, line 22: related names are missing

db/models.py
```diff
@@ -14,3 +14,43 @@ class Actor(models.Model):
...
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(
+        CinemaHall,
+        on_delete=models.CASCADE,
+        related_name="sessions"
+    )
+    movie = models.ForeignKey(
+        Movie,
+        on_delete=models.CASCADE,
+        related_name="movies"
```
Oleksl888, line 52: think of a better related name here

services/movie.py
```diff
@@ -0,0 +1,38 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+    genres_ids: list = None,
```
Oleksl888, line 7: Annotate optional parameters - check in other functions too

services/movie.py
```diff
@@ -0,0 +1,38 @@
...
+
+    if genres_ids:
+        queryset = queryset.filter(genres__id__in=genres_ids).distinct()
+    if actors_ids:
+        queryset = queryset.filter(actors__id__in=actors_ids).distinct()
+    return queryset
+
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
```
Oleksl888, line 20: using pk is preferable in django

services/movie_session.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def create_movie_session(
+        movie_show_time: datetime,
+        movie_id: int,
+        cinema_hall_id: int,
+) -> MovieSession:
+    new_movie_session = MovieSession.objects.create(
+        show_time=movie_show_time,
+        cinema_hall=CinemaHall.objects.get(id=cinema_hall_id),
```
Oleksl888, line 15: no need to pass objects here - you can pass ids directly

services/movie_session.py
```diff
@@ -0,0 +1,57 @@
...
+        movie=Movie.objects.get(id=movie_id)
+    )
+    return new_movie_session
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
+    if session_date:
+        target_time = datetime.strptime(session_date, "%Y-%m-%d").date()
```
Oleksl888, line 23: Can you refactor this to use only one return?
Also you may pass string as is without conversion - django will take care of everything under the hood

services/movie_session.py
```diff
@@ -0,0 +1,57 @@
...
+
+    if show_time:
+        movie_session.show_time = show_time
+
+    if cinema_hall_id:
+        cinema_hall_inst = CinemaHall.objects.filter(id=cinema_hall_id)
+        if cinema_hall_inst.exists():
+            movie_session.cinema_hall = cinema_hall_inst.first()
+
+    if movie_id:
```
Oleksl888, line 48: Can you refactor this so that you will not need to get movie session if none of the update parameters has been passed?

services/movie_session.py
```diff
@@ -0,0 +1,57 @@
...
+        movie_inst = Movie.objects.filter(id=movie_id)
+        if movie_inst.exists():
+            movie_session.movie = movie_inst.first()
+
+    movie_session.save()
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    MovieSession.objects.filter(id=session_id).delete()
```
Oleksl888, line 57: Here you could reuse `get_movie_session_by_id` function to make code DRY

Oleksl888 (changes requested):

services/movie_session.py
ihorutkin: When I  did it i get an error:  ValueError: Cannot assign "2": "MovieSession.cinema_hall" must be a "CinemaHall" instance.

services/movie_session.py
ihorutkin: Excuse me, can you please explain this note to me as I am getting the movie session only if some parameters are passed

services/cinema_hall.py
```diff
@@ -0,0 +1,20 @@
...
+        hall_name: str,
+        hall_rows: int,
+        hall_seats_in_row: int
+) -> CinemaHall:
+    new_cinema_hall = CinemaHall.objects.create(
+        name=hall_name,
+        rows=hall_rows,
+        seats_in_row=hall_seats_in_row
+    )
+    return new_cinema_hall
```
oliasymak24, line 20: You don’t need to `return` anything in create/update/delete methods, check everywhere

services/movie.py
```diff
@@ -0,0 +1,38 @@
...
+    genres_ids: list[int] = None,
+    actors_ids: list[int] = None
+) -> QuerySet:
+    queryset = Movie.objects.all()
+
+    if genres_ids:
+        queryset = queryset.filter(genres__id__in=genres_ids).distinct()
+    if actors_ids:
+        queryset = queryset.filter(actors__id__in=actors_ids).distinct()
+    return queryset
```
oliasymak24, line 16: Better to use distinct ony once, here

services/movie_session.py
```diff
@@ -0,0 +1,58 @@
...
+        cinema_hall_inst = CinemaHall.objects.filter(id=cinema_hall_id)
+        if cinema_hall_inst.exists():
+            movie_session.cinema_hall = cinema_hall_inst.first()
+
+    if movie_id:
+        movie_inst = Movie.objects.filter(id=movie_id)
+        if movie_inst.exists():
+            movie_session.movie = movie_inst.first()
+
+    movie_session.save()
```
oliasymak24, line 54: Try to use updte function

oliasymak24 (changes requested):

services/movie.py
```diff
@@ -0,0 +1,37 @@
...
+    genres_ids: list[int] = None,
+    actors_ids: list[int] = None
+) -> QuerySet:
+    queryset = Movie.objects.all()
+
+    if genres_ids:
+        queryset = queryset.filter(genres__id__in=genres_ids)
+    if actors_ids:
+        queryset = queryset.filter(actors__id__in=actors_ids)
+    return queryset.distinct()
```
MNDonut, line 16: 1 blank line after the last if statement

services/movie.py
```diff
@@ -0,0 +1,37 @@
...
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(pk=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
```
MNDonut, line 27: list or None type

services/movie_session.py
MNDonut: It's because you have explicit field cinema_hall and movie(that you added in the models). But under the hood in database it's saved as cinema_hall_id and movie_id because database doesn't know what is Django and related models. So you can use these "hidden" fields and set id instead of making redundant queries
@ihorutkin

services/movie_session.py
```diff
@@ -0,0 +1,55 @@
...
+        movie=Movie.objects.get(id=movie_id)
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
+    queryset = MovieSession.objects.all()
+    if session_date:
+        target_time = datetime.strptime(session_date, "%Y-%m-%d")
+        queryset = queryset.filter(show_time__date=target_time)
+    return queryset
```
MNDonut, line 25: blank line after if

services/movie_session.py
```diff
@@ -0,0 +1,55 @@
...
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(pk=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: datetime = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
```
MNDonut, line 36: int/datetime or None type

services/movie_session.py
```diff
@@ -0,0 +1,55 @@
...
+) -> None:
+    movie_session = MovieSession.objects.filter(id=session_id)
+
+    if show_time:
+        movie_session.update(show_time=show_time)
+
+    if cinema_hall_id:
+        cinema_hall_inst = CinemaHall.objects.filter(id=cinema_hall_id)
+        if cinema_hall_inst.exists():
```
MNDonut, line 45: no need to fetch the object for the reason I explained above. Use hidden field instead

MNDonut (changes requested):

services/movie.py
ihorutkin: Can you explain me please this note, I wrote None cuz it's just default value

services/movie_session.py
ihorutkin: Here also)

db/models.py
```diff
@@ -14,3 +14,43 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="actors")
```
Oleksl888, line 22: this is the name that will refer to related `movies` on Actor table - so why would you call it so?

db/models.py
```diff
@@ -14,3 +14,43 @@ class Actor(models.Model):
...
+    genres = models.ManyToManyField(Genre, related_name="genres")
```
Oleksl888, line 23: change here as well please

db/models.py
```diff
@@ -14,3 +14,43 @@ class Actor(models.Model):
...
+    def __str__(self) -> str:
+        return self.name
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(
+        CinemaHall,
+        on_delete=models.CASCADE,
+        related_name="cinema_halls"
```
Oleksl888, line 47: this is a misleading name as well - revise documentation or learning materials about related names

services/movie.py
```diff
@@ -0,0 +1,38 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+    genres_ids: list[int] = None,
```
Oleksl888, line 7: these should be annotated as optional parameters

services/movie.py
Oleksl888: Right - when you assign default value as None it means that parameter is optional but it is not explicitly shown in type annotation. You should use `Optional` (old school style) or use `| None` ( as of python 3.9) on `list[int]` . Please check and adjust everywhere throughout your code

services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+        show_time=movie_show_time,
+        cinema_hall_id=cinema_hall_id,
+        movie_id=movie_id,
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
+    queryset = MovieSession.objects.all()
+    if session_date:
+        target_time = datetime.strptime(session_date, "%Y-%m-%d")
```
Oleksl888, line 23: you can pass date without conversion - django will take of it under the hood

services/movie_session.py
Oleksl888: Try to implement this:
1. Create an empty dict
2. Check if any of the optional arguments are passed to this function. In case if there are - you will add them to the dict.
3.  In the end check if your dict has some values to update and if yes you will call update on queryset, and if there are not any you don't do anything.

Oleksl888 (changes requested):

services/movie.py
ihorutkin: Ok, got you, now I understand

db/models.py
```diff
@@ -14,3 +14,43 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="movies_actors")
```
Oleksl888, line 22: it must be `movies` because it is referring `Movies` from `Actors` table

db/models.py
```diff
@@ -14,3 +14,43 @@ class Actor(models.Model):
...
+    genres = models.ManyToManyField(Genre, related_name="movies_genres")
```
Oleksl888, line 23: here must be `movies` as well

db/models.py
```diff
@@ -14,3 +14,43 @@ class Actor(models.Model):
...
+    def __str__(self) -> str:
+        return self.name
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(
+        CinemaHall,
+        on_delete=models.CASCADE,
+        related_name="hall_sessions"
```
Oleksl888, line 47: it must be `movie_sessions` here too

services/movie_session.py
```diff
@@ -0,0 +1,63 @@
...
+        show_time=movie_show_time,
+        cinema_hall_id=cinema_hall_id,
+        movie_id=movie_id,
+    )
+
+
+def get_movies_sessions(session_date: str | None = None) -> QuerySet:
+    queryset = MovieSession.objects.all()
+    if session_date:
+        target_time = session_date
```
Oleksl888, line 23: you don't need this assignment

services/movie_session.py
Oleksl888: almost done it:
remove line 39 where you get movie session and just put
`MovieSession.objects.filter(pk=session_id).update(**updating_data)` 
under `if updating_data:`

Oleksl888 (changes requested):

Oleksl888 (changes requested):
you haven't changed anything

LiudmylaKulinchenko (approved):
Good job 🌟🤠

PR #927, opened 2024-08-28
Solution

pashawarganov:


db/models.py
```diff
@@ -14,3 +14,37 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="actors_movies")
```
Oleksl888, line 22: `movies` would be more appropriate name, since the model is called Movie

db/models.py
```diff
@@ -14,3 +14,37 @@ class Actor(models.Model):
...
+        return self.name
+
+    @property
+    def capacity(self) -> int:
+        return self.rows * self.seats_in_row
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall, on_delete=models.CASCADE,
```
Oleksl888, line 44: Think of a better related name + place arguments each on one line

services/movie.py
```diff
@@ -0,0 +1,41 @@
...
+    if genres_ids:
+        movies = movies.filter(genres__id__in=genres_ids)
+    if actors_ids:
+        movies = movies.filter(actors__id__in=actors_ids)
+
+    return movies
+
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
```
Oleksl888, line 21: using pk is preferable in django

services/movie_session.py
```diff
@@ -0,0 +1,56 @@
...
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: Optional[datetime] = None,
+        movie_id: Optional[int] = None,
+        cinema_hall_id: Optional[int] = None
+) -> MovieSession:
+    updated_movie_session = MovieSession.objects.get(id=session_id)
```
Oleksl888, line 39: Can you refactor this so that you will not need to get movie session if none of the update parameters has been passed?

services/movie_session.py
```diff
@@ -0,0 +1,56 @@
...
+    if cinema_hall_id:
+        updated_movie_session.cinema_hall_id = cinema_hall_id
+
+    updated_movie_session.save()
+
+    return updated_movie_session
+
+
+def delete_movie_session_by_id(session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=session_id).delete()
```
Oleksl888, line 56: Here you could reuse get_movie_session_by_id function to make code DRY. Also double check the return type of delete() method

settings.py
```diff
@@ -21,6 +21,6 @@
 
 USE_I18N = True
 
-USE_TZ = False
+USE_TZ = True
```

Oleksl888 (changes requested):

services/movie.py
pashawarganov: got it

MNDonut (approved):

PR #928, opened 2024-08-28
Solution

Bogdan-Rylov:


services/movie.py
```diff
@@ -0,0 +1,45 @@
...
+from db.models import Genre, Actor, Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
+) -> QuerySet:
+    movies = Movie.objects.all()
+
+    if genres_ids and actors_ids:
```
Oleksl888, line 12: you don't need this condition two `if` statements should be enough for this function to produce correct result

Oleksl888 (changes requested):

services/movie_session.py
```diff
@@ -0,0 +1,54 @@
...
+        movie_session = MovieSession.objects.get(pk=session_id)
+
+        if show_time:
+            movie_session.show_time = show_time
+        if movie_id:
+            movie_session.movie_id = movie_id
+        if cinema_hall_id:
+            movie_session.cinema_hall_id = cinema_hall_id
+
+        movie_session.save()
```
oliasymak24, line 50: Try to use update function here

services/movie.py
```diff
@@ -0,0 +1,43 @@
...
+        genres_ids: Optional[list[int]] = None,
+        actors_ids: Optional[list[int]] = None
+) -> None:
+    movie = Movie.objects.create(
+        title=movie_title,
+        description=movie_description
+    )
+
+    if genres_ids:
+        genres = Genre.objects.filter(id__in=genres_ids)
```
oliasymak24, line 38: No need to add this filtering

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #929, opened 2024-08-29
implement (first attempt)

Maksym-Turenko:


db/migrations/0002_cinemahall_movie_moviesession.py
oliasymak24, line 1: Where is your 0001 file? :)

oliasymak24 (changes requested):

db/migrations/0002_cinemahall_movie_moviesession.py
Maksym-Turenko: This is the default file I get when cloning a project from GitHub, and since I haven't changed it, I can't push it.

oliasymak24 (approved):

PR #931, opened 2024-08-30
Solution for movie session

wataschi:


db/migrations/0002_cinemahall_movie.py
```diff
@@ -0,0 +1,32 @@
+# Generated by Django 4.0.2 on 2024-08-30 13:29
```
pavlopro, line 1: It is a good practice to include one migration per app per PR in Django projects since it helps to avoid migrations folder overload in the future. Represent all changes done to models in one migration file per app per PR, please

services/cinema_hall.py
```diff
@@ -0,0 +1,19 @@
+from typing import List
+
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> List[CinemaHall]:
```
pavlopro, line 6: Please built-in list instead of `List`

services/movie.py
```diff
@@ -0,0 +1,35 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None,
+) -> QuerySet:
```
pavlopro, line 9: Indicate QuerySet of which instances you expect

services/movie_session.py
```diff
@@ -0,0 +1,49 @@
+from db.models import MovieSession
+
+import datetime
+
```
pavlopro, line 4: Please correct import groups ordering according to pep8

services/movie_session.py
```diff
@@ -0,0 +1,49 @@
...
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: datetime.datetime = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> None:
+    movie_session = MovieSession.objects.get(id=session_id)
```
pavlopro, line 38: use `get_movie_session_by_id `method here

pavlopro (changes requested):

services/movie_session.py
```diff
@@ -0,0 +1,52 @@
...
+from django.db.models import QuerySet
+
+from db.models import MovieSession
+
+
+def create_movie_session(
+        movie_show_time: datetime.datetime,
+        movie_id: int,
+        cinema_hall_id: int,
+) -> None:
```
LLkaia, line 12: But it returns MovieSession

LLkaia (changes requested):

services/movie_session.py
wataschi: My bad, i fixed it

LLkaia (approved):

PR #939, opened 2024-09-10
Solution

EugeneDolinskiy:


db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
...
+    @property
+    def capacity(self) -> int:
+        return self.seats_in_row * self.rows
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(CinemaHall, on_delete=models.CASCADE)
+    movie = models.ForeignKey(Movie, on_delete=models.CASCADE)
+
```
pavlopro, line 46: Please add related names for relational fields

services/cinema_hall.py
```diff
@@ -0,0 +1,19 @@
+from django.db.models import QuerySet
+
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> QuerySet:
```
pavlopro, line 6: Indicate QuerySet if which instances you expect

services/movie_session.py
```diff
@@ -0,0 +1,51 @@
...
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: datetime = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> None:
+    movie_session = MovieSession.objects.get(id=session_id)
```
pavlopro, line 38: use `get_movie_session_by_id` method

pavlopro (changes requested):

db/migrations/0002_cinemahall_movie_moviesession.py
```diff
@@ -0,0 +1,42 @@
+# Generated by Django 4.0.2 on 2024-09-10 13:54
+
```
pavlopro, line 2: It is a good practice to include one migration per app per PR in Django projects since it helps to avoid migrations folder overload in the future. Represent all changes done to models in one migration file per app per PR, please

pavlopro (changes requested):

pavlopro (changes requested):
Do not delete existing migration file.
You should have one migration file per app per PR. So you just need to create only one new migration file

PR #940, opened 2024-09-12
Develop

comnotaDidi:


db/migrations/models.py
Dimosphen1, line 1: There is no need to place this file inside the `migrations` folder. Create and add a single migration here instead

db/models.py
Dimosphen1, line 1: There is no need to delete this file, leave it as is and extend its content according to the task description

Dimosphen1 (changes requested):
Several changes were requested. Make sure that all tests pass

vsmutok (changes requested):
Tests Failed!

PR #942, opened 2024-09-13
Solution

frezworx:


db/migrations/0002_movie.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 4.0.2 on 2024-09-11 09:10
```
pavlopro, line 1: It is a good practice to include one migration per app per PR in Django projects since it helps to avoid migrations folder overload in the future. Represent all changes done to models in one migration file per app per PR, please

services/movie.py
```diff
@@ -0,0 +1,45 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None,
+) -> list[Movie] | QuerySet[Movie]:
```
pavlopro, line 9: remove this `list[Movie]`

services/movie.py
```diff
@@ -0,0 +1,45 @@
...
+    movies = Movie.objects.all()
+    if genres_ids and actors_ids:
+        movies = movies.filter(
```
pavlopro, line 12:
```
if genres_ids and actors_ids:
...
```
This block of code is redundant. Remvoe it

services/movie.py
```diff
@@ -0,0 +1,45 @@
...
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
+) -> Movie:
+    movie, _ = Movie.objects.get_or_create(
```
pavlopro, line 34: use `create` instead of `get_or_create`

services/movie.py
```diff
@@ -0,0 +1,45 @@
...
+        description=movie_description,
+    )
+
+    if genres_ids:
+        movie.genres.set(genres_ids)
+    if actors_ids:
+        movie.actors.set(actors_ids)
+
+        movie.save()
+        return movie
```
pavlopro, line 45: This is incorrect. It shouldn't be nested into `if` statement

services/movie_session.py
```diff
@@ -0,0 +1,61 @@
...
+        show_time=movie_show_time,
+        movie=movie,
+        cinema_hall=cinema_hall
+    )
+
+
+def get_movies_sessions(
+        session_date: str = None
+) -> QuerySet:
+    if session_date:
```
pavlopro, line 26:
```
if session_date:
        return MovieSession.objects.filter(
            show_time__date=session_date
        )
return MovieSession.objects.all()
```

services/movie_session.py
```diff
@@ -0,0 +1,61 @@
...
+    MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie=movie,
+        cinema_hall=cinema_hall
+    )
+
+
+def get_movies_sessions(
+        session_date: str = None
+) -> QuerySet:
```
pavlopro, line 25: indicate QuerySet of which instances you expect

services/movie_session.py
```diff
@@ -0,0 +1,61 @@
...
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: datetime = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> None:
+    session = MovieSession.objects.get(id=session_id)
```
pavlopro, line 45: use `get_movie_session_by_id` method

services/movie_session.py
```diff
@@ -0,0 +1,61 @@
...
+
+    if show_time:
+        session.show_time = show_time
+    if movie_id:
+        movie = Movie.objects.get(id=movie_id)
```
pavlopro, line 50: instead of `movie = Movie.objects.get(id=movie_id)`
use `session.movie_id = movie_id`

services/movie_session.py
```diff
@@ -0,0 +1,61 @@
...
+        session.movie = movie
+    if cinema_hall_id:
+        cinema_hall = CinemaHall.objects.get(id=cinema_hall_id)
+        session.cinema_hall = cinema_hall
```
pavlopro, line 54: here the same

services/movie_session.py
```diff
@@ -0,0 +1,61 @@
...
+
+    session.save()
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    session = MovieSession.objects.get(id=session_id)
```
pavlopro, line 60: No need to have this variable 
simplify to `get_movie_session_by_id(id=session_id).delete()`

pavlopro (changes requested):

PR #971, opened 2024-10-23
created models and services

Mxbely:


db/models.py
```diff
@@ -14,3 +14,41 @@ class Actor(models.Model):
...
+    cinema_hall = models.ForeignKey(CinemaHall,
+                                    on_delete=models.CASCADE,
+                                    related_name="movie_sessions")
+    movie = models.ForeignKey(Movie, on_delete=models.DO_NOTHING,
+                              related_name="movie_sessions")
+
+    def __str__(self) -> str:
+        return (
+            f"{self.movie.title} "
+            f'{self.show_time.strftime("%Y-%m-%d %H:%M:%S")}'  # noqa: E999
```
vladyslav-tmf, line 53: You can use single quotes inside the f-string, then flake8 will not raise an error.
```suggestion
            f"{self.movie.title} "
            f"{self.show_time.strftime('%Y-%m-%d %H:%M:%S')}"

File: services/movie.py [hunk 2]
```diff
@@ -0,0 +1,29 @@
...
+               actors_ids: list[int] = None) -> QuerySet:
+    set_ = Movie.objects.all()
+    if genres_ids:
+        set_ = set_.filter(genres__id__in=genres_ids)
+    if actors_ids:
+        set_ = set_.filter(actors__id__in=actors_ids)
+    return set_
+
+
+def get_movie_by_id(movie_id: int = None) -> Movie:
```
vladyslav-tmf, line 16: `None` is not a valid value for retrieving a record from the database. If you don't pass movie_id, the method will try to execute Movie.objects.get(id=None), which will result in an error because None is not a valid id.

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+
+from db.models import MovieSession, CinemaHall, Movie
+
+
+def create_movie_session(movie_show_time: datetime,
+                         movie_id: int, cinema_hall_id: int) -> MovieSession:
+    return MovieSession.objects.create(
+        show_time=movie_show_time,
+        cinema_hall=CinemaHall.objects.get(id=cinema_hall_id),
+        movie=Movie.objects.get(id=movie_id)
```
vladyslav-tmf, line 13: As checklist says:
Don't load the DB when it's not needed.

Before creating a movie session, an additional database query is performed to retrieve the CinemaHall and Movie objects by their id. This is an additional load on the database because you are loading objects twice when you don't need to.
Django allows you to use `movie_id` and `cinema_hall_id` directly when creating an object using a ForeignKey. In this case, you don't have to load extra objects from the database, which makes this approach more efficient.

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet | MovieSession:
```
vladyslav-tmf, line 17: Actually, the method always returns a `QuerySet`, not `MovieSession`.

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> None:
+    session = MovieSession.objects.get(id=session_id)
+    if show_time:
+        session.show_time = show_time
+    if movie_id:
+        session.movie = Movie.objects.get(id=movie_id)
+    if cinema_hall_id:
+        session.cinema_hall = CinemaHall.objects.get(id=cinema_hall_id)
```
vladyslav-tmf, line 45: Same here: Don't load the DB when it's not needed.

vladyslav-tmf (changes requested):
In general, the code looks good)
You just need to fix a few minor issues.

db/models.py
Mxbely: When i use it, i also have flake8 error:
`Q000 Single quotes found but double quotes preferred`

services/movie_session.py
Mxbely: if you watched videos, you can remember, that use direct access to ForeginKey is a bad practice, and use .get() is a good practice

services/movie_session.py
Mxbely: Same: use .get() is a good practice

Nattalli: Disagree with you. Read the checklist (code efficiency): https://github.com/mate-academy/py-movie-moviesession-cinemahall/blob/master/checklist.md#code-efficiency

Nattalli: so please, fix it

services/movie_session.py
```diff
@@ -0,0 +1,46 @@
...
+        cinema_hall=CinemaHall.objects.get(id=cinema_hall_id),
+        movie=Movie.objects.get(id=movie_id)
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
+    set_ = MovieSession.objects.all()
+    if session_date:
+        dat = datetime.datetime.strptime(session_date, "%Y-%m-%d")
+        set_ = set_.filter(show_time__date=dat)
```
Nattalli, line 21: bad naming

services/movie.py
```diff
@@ -0,0 +1,30 @@
...
+    if movie_id:
+        return Movie.objects.get(id=movie_id)
+
+
+def create_movie(movie_title: str, movie_description: str,
+                 genres_ids: list[int] = None,
+                 actors_ids: list[int] = None) -> Movie:
+    new_movie = Movie.objects.create(title=movie_title,
+                                     description=movie_description,)
```
Nattalli, line 25:
```suggestion
    new_movie = Movie.objects.create(title=movie_title, description=movie_description)
```

services/cinema_hall.py
```diff
@@ -0,0 +1,16 @@
...
+
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> QuerySet:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(hall_name: str, hall_rows: int,
+                       hall_seats_in_row: int) -> CinemaHall:
```
Nattalli, line 11:
```suggestion
def create_cinema_hall(
     hall_name: str, 
     hall_rows: int,
     hall_seats_in_row: int
) -> CinemaHall:
```

Nattalli (changes requested):

services/movie.py
Mxbely: flake8 will not pass

services/movie.py
```diff
@@ -0,0 +1,32 @@
...
+    if genres_ids:
+        set_ = set_.filter(genres__id__in=genres_ids)
+    if actors_ids:
+        set_ = set_.filter(actors__id__in=actors_ids)
+    return set_
+
+
+def get_movie_by_id(movie_id: int = None) -> Movie:
+    if movie_id:
+        return Movie.objects.get(id=movie_id)
```
AnyoneClown, line 18: You don't need to use default value here. U should expect to retrieve proper `id` here

services/movie.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def create_movie(movie_title: str, movie_description: str,
+                 genres_ids: list[int] = None,
+                 actors_ids: list[int] = None) -> Movie:
```
AnyoneClown, line 23: Start every argument from a new line

services/movie_session.py
```diff
@@ -0,0 +1,46 @@
+import datetime
+
+from django.db.models import QuerySet
+
+from db.models import MovieSession
+
+
+def create_movie_session(movie_show_time: datetime,
+                         movie_id: int, cinema_hall_id: int) -> MovieSession:
```
AnyoneClown, line 9: Same here

services/movie_session.py
```diff
@@ -0,0 +1,46 @@
...
+        cinema_hall_id=cinema_hall_id,
+        movie_id=movie_id,
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
+    queryset = MovieSession.objects.all()
+    if session_date:
+        dat = datetime.datetime.strptime(session_date, "%Y-%m-%d")
+        queryset = queryset.filter(show_time__date=dat)
```
AnyoneClown, line 21: Use correct variable name

services/movie_session.py
```diff
@@ -0,0 +1,46 @@
...
+    return MovieSession.objects.get(id=movie_session_id)
+
+
+def update_movie_session(
+        session_id: int,
+        show_time: datetime = None,
+        movie_id: int = None,
+        cinema_hall_id: int = None
+) -> None:
+    session = MovieSession.objects.get(id=session_id)
```
AnyoneClown, line 35: Reuse `get_movie_session_by_id`

services/movie_session.py
```diff
@@ -0,0 +1,46 @@
...
+        session.show_time = show_time
+    if movie_id:
+        session.movie_id = movie_id
+    if cinema_hall_id:
+        session.cinema_hall_id = cinema_hall_id
+    session.save()
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    MovieSession.objects.filter(id=session_id).delete()
```
AnyoneClown, line 46: Reuse `get_movie_session_by_id`

AnyoneClown (changes requested):

Arsen-hrynevych (approved):
ok

PR #973, opened 2024-10-24
Solution

vladyslav-tmf:


db/models.py
```diff
@@ -14,3 +14,39 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="movies")
+    genres = models.ManyToManyField(Genre, related_name="movies")
```
AnyoneClown, line 23: Related name should probably be `genres` here.

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+) -> MovieSession:
+    return MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
+
+
+def get_movies_sessions(
+    session_date: str = None
```
AnyoneClown, line 19: `session_date` should be `date` type, not `str`. Fix in other places too

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+        movie_session.movie_id = movie_id
+    if cinema_hall_id:
+        movie_session.cinema_hall_id = cinema_hall_id
+
+    movie_session.save()
+    return movie_session
+
+
+def delete_movie_session_by_id(session_id: int) -> None:
+    MovieSession.objects.get(id=session_id).delete()
```
AnyoneClown, line 50: Reuse `get_movie_session_by_id` here

AnyoneClown (changes requested):

PR #979, opened 2024-10-29
Solution

KKvladys:


db/models.py
```diff
@@ -14,3 +14,39 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="actors")
+    genres = models.ManyToManyField(Genre, related_name="movies")
```
vladyslav-tmf, line 23: Use the correct `related_name`. In the majority of cases, the name of the model in plural will be just right.

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+
+def create_movie_session(
+        movie_show_time: datetime,
+        movie_id: int,
+        cinema_hall_id: int
+) -> MovieSession:
+    return MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie=Movie.objects.get(id=movie_id),
+        cinema_hall=CinemaHall.objects.get(id=cinema_hall_id)
```
vladyslav-tmf, line 15: Don't load the DB when it's not needed.
This function can be improved by avoiding unnecessary database queries. Consider using the `movie_id` and `cinema_hall_id` directly in the create method instead of fetching the objects first.

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
+    queryset = MovieSession.objects.all()
+
+    if session_date:
+        queryset = queryset.filter(show_time__date=session_date)
+    return queryset
```
vladyslav-tmf, line 24: This function unnecessarily loads the database with `MovieSession.objects.all()` even when `session_date` is not provided.
The function should only load `MovieSession.objects.all()` if `session_date` is provided as `None`.

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+        cinema_hall_id: int = None
+) -> None:
+    movie_session = get_movie_session_by_id(session_id)
+
+    if show_time:
+        movie_session.show_time = show_time
+    if movie_id:
+        movie_session.movie = Movie.objects.get(id=movie_id)
+    if cinema_hall_id:
+        movie_session.cinema_hall = CinemaHall.objects.get(id=cinema_hall_id)
```
vladyslav-tmf, line 44: Don't load the DB when it's not needed, same here.

PR #985, opened 2024-11-04
Solution

MatvijSafonov:


services/movie_session.py
```diff
@@ -0,0 +1,44 @@
...
+    if movie_id:
+        movie_session.movie_id = movie_id
+    if cinema_hall_id:
+        movie_session.cinema_hall_id = cinema_hall_id
+    movie_session.save()
+    return movie_session
+
+
+def delete_movie_session_by_id(session_id: None) -> None:
+    MovieSession.objects.filter(id=session_id).delete()
```
Atikiho, line 44: You should use `.get `when there is one object to get. Here is quote from checklist

> Don't use .filter() when you need to get one item, use .get() instead.
>
> Good example:
> 
```
player_id = 1

Player.objects.get(id=player_id)
```

> Bad example:
```
player_id = 1

Player.objects.filter(id=player_id)
```

Atikiho (approved):

services/movie_session.py
MatvijSafonov: Thx for noticing this, allmost forgot about it

PR #998, opened 2024-11-06
solution

serhiik05:


services/movie.py
```diff
@@ -0,0 +1,36 @@
...
+               ) -> QuerySet | Movie:
+    if genres_ids is None and actors_ids is None:
+        return Movie.objects.all()
+    if genres_ids and actors_ids:
+        return Movie.objects.filter(genres__id__in=genres_ids,
+                                    actors__id__in=actors_ids).distinct()
+    if genres_ids and (actors_ids is None):
+        return Movie.objects.filter(genres__id__in=genres_ids)
+    if actors_ids and (genres_ids is None):
+        return Movie.objects.filter(actors__id__in=actors_ids)
```
KKvladys, line 17: maybe it will be better here write all Movie in variable and then check the conditions one by one and filter the variable

KKvladys (approved):

PR #1007, opened 2024-11-06
solution

dxrrkwm:


services/movie.py
```diff
@@ -0,0 +1,45 @@
...
+def create_movie(movie_title: str,
+                 movie_description: str,
+                 genres_ids: list[int] = None,
+                 actors_ids: list[int] = None) -> Movie:
+
+    movie = Movie.objects.create(
+        title=movie_title,
+        description=movie_description
+    )
+
```
dmytrik, line 37: you can use method set, that don't make unnecessary requests(movie.genres.set(genres_ids))

dmytrik (approved):

PR #1008, opened 2024-11-07
Solution

FursykIvan:


services/movie.py
```diff
@@ -0,0 +1,38 @@
...
+    movies = Movie.objects.all()
+
+    if genres_ids and actors_ids:
+        movies = movies.filter(genres__id__in=genres_ids,
+                               actors__id__in=actors_ids)
+    elif genres_ids:
+        movies = movies.filter(genres__id__in=genres_ids)
+    elif actors_ids:
+        movies = movies.filter(actors__id__in=actors_ids)
+
```
dmytrik, line 15: you don't need first check

dmytrik (approved):

services/movie.py
FursykIvan: Thank you!)

PR #1009, opened 2024-11-07
Solution

sofiiasavkova:


services/movie.py
```diff
@@ -0,0 +1,35 @@
...
+        actors_ids: list[int]
+) -> Movie:
+    movie = Movie.objects.create(
+        title=movie_title,
+        description=movie_description
+    )
+    if genres_ids:
+        movie.genres.set(Genre.objects.filter(id__in=genres_ids))
+    if actors_ids:
+        movie.actors.set(Actor.objects.filter(id__in=actors_ids))
```
94nj111, line 34: __set__ can work with ids

services/movie_session.py
```diff
@@ -0,0 +1,50 @@
...
+        movie_show_time: datetime,
+        movie_id: int,
+        cinema_hall_id: int
+) -> MovieSession:
+    movie = Movie.objects.get(id=movie_id)
+    cinema_hall = CinemaHall.objects.get(id=cinema_hall_id)
+    return MovieSession.objects.create(
+        show_time=movie_show_time,
+        movie=movie,
+        cinema_hall=cinema_hall
```
94nj111, line 16: you can use movie_id and cinema_hall_id to avoid getting objects from db

cinema_db_data.json
```diff
@@ -259,7 +259,7 @@
     "model": "db.moviesession",
     "pk": 4,
     "fields": {
-      "show_time": "2024-10-11T13:00:00Z",
+      "show_time": "2024-10-11T13:00:00",
       "movie": 4,
       "cinema_hall": 4
     }
```
94nj111, line 265: no need to change it, the tests will still pass

94nj111 (changes requested):
and don't forget to add the migration file

94nj111 (approved):
Good job!

PR #1066, opened 2025-01-09
Solution

IvankaKuzin:


services/movie.py
```diff
@@ -0,0 +1,39 @@
...
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: int = None,
+        actors_ids: int = None,
+) -> Movie:
```
julia4406, line 29: genres_ids and actors_ids are list[int]

services/movie_session.py
```diff
@@ -0,0 +1,76 @@
...
+            movie_id = Movie.objects.get(name=movie_id)
+
+    new_movie_session = MovieSession.objects.create(
+        show_time=movie_show_time,
+        cinema_hall=cinema_hall_id,
+        movie=movie_id,
+    )
+
+    return new_movie_session
+
```
julia4406, line 31: you don't need to return new session, just create

julia4406 (approved):
Cooool

services/movie_session.py
```diff
@@ -0,0 +1,76 @@
...
+from django.db.models import QuerySet
+
+from db.models import MovieSession, Movie, CinemaHall
+
+
+def create_movie_session(
+        movie_show_time: datetime,
+        movie_id: str,
+        cinema_hall_id: str
```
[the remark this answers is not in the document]
petrykivd: Agree with AI mentor

services/movie.py
```diff
@@ -0,0 +1,39 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list = None,
+        actors_ids: list = None
+) -> list | QuerySet:
```
petrykivd, line 9: Stick to the same code style, if you specify list[int] below, you should also specify it here

petrykivd (approved):
Nice! 👍

PR #1071, opened 2025-01-10
solution_v.1.0

yelizaveta-zh:


db/models.py
```diff
@@ -14,3 +14,33 @@ class Actor(models.Model):
...
+
+    def __str__(self) -> str:
+        return self.title
+
+
+class CinemaHall(models.Model):
+    name = models.CharField(max_length=255)
+    row = models.IntegerField()
+    seat_count = models.IntegerField()
+
```
Dogober, line 33: The correct name for this field is = seats_in_row. According to README.md

db/models.py
```diff
@@ -14,3 +14,33 @@ class Actor(models.Model):
...
+    def capacity(self):
+        return range(1, self.seat_count + 1)
```
Dogober, line 35: This func returns range obj but you need to return exactly int:
blue = CinemaHall.objects.create(name="Blue", rows=9, seats_in_row= 13) 

print(blue.capacity)
output -> 117

db/models.py
```diff
@@ -14,3 +14,33 @@ class Actor(models.Model):
...
+    def __str__(self) -> str:
+        return self.name
+
+
+class MovieSession(models.Model):
+    movie = models.ForeignKey(Movie, on_delete=models.CASCADE)
+    date = models.DateField()
+    time = models.TimeField()
+    price = models.DecimalField(max_digits=6, decimal_places=2)
+
```
Dogober, line 46: You need to rewrite into the coorect form:
MovieSession, with such field:
date time field **show_time**, the date and the time of the movie session performance
foreign key **cinema_hall**, the hall where the movie session is performed, references to the table CinemaHall
foreign key **movie**, the movie to be shown, references to the table Movie

Dogober (changes requested):
Also you need to implement the rest logic from task

db/models.py
```diff
@@ -12,5 +12,37 @@ class Actor(models.Model):
     first_name = models.CharField(max_length=255)
     last_name = models.CharField(max_length=255)
 
+    def __str__(self) -> tuple[str, str]:
+        return self.first_name, self.last_name
```
petrykivd, line 16: Maybe it's better to return string?

services/movie.py
```diff
@@ -0,0 +1,49 @@
...
+    if genres_ids and actors_ids:
+        movies = movies.filter(
+            Q(genres__id__in=genres_ids) & Q(actors__id__in=actors_ids)
+        ).distinct()
+    elif genres_ids:
+        movies = movies.filter(genres__id__in=genres_ids).distinct()
+    elif actors_ids:
+        movies = movies.filter(actors__id__in=actors_ids).distinct()
+
+    return movies
```
petrykivd, line 22: If you want to use .distinct(), maybe it's better to use it once at the end?

services/movie_session.py
```diff
@@ -0,0 +1,51 @@
...
+        cinema_hall_id=cinema_hall_id
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
+
+    if session_date:
+        return MovieSession.objects.filter(show_time__date=session_date)
+    else:
+        return MovieSession.objects.all()
```
petrykivd, line 25: Simplify it

petrykivd (changes requested):

petrykivd (approved):
Well done! 🚀

PR #1082, opened 2025-01-12
Solution

Professor-Douel:


services/movie_session.py
```diff
@@ -0,0 +1,64 @@
...
+        show_time=datetime.strptime(
+            movie_show_time,
+            "%Y-%m-%d %H:%M:%S"
+        ),
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id
+    )
+
+
+def get_movie_sessions(session_date: Optional[str]) -> "List[MovieSession]":
```
taraskhom, line 22: Tests can't find this method because you need to name it "get_movies_sessions" instead of "get_movie_sessions"

taraskhom (changes requested):

PR #1143, opened 2025-02-24
Solution

JohnyJDX:


db/models.py
```diff
@@ -14,3 +14,45 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField(blank=True, null=True)
+    actors = models.ManyToManyField(Actor, related_name="movies")
+    genres = models.ManyToManyField(Genre, related_name="genres")
```
Nattalli, line 23: better name it `movies`:
```suggestion
    genres = models.ManyToManyField(Genre, related_name="movies")
```

services/cinema_hall.py
```diff
@@ -0,0 +1,18 @@
+from django.db.models import QuerySet
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> QuerySet:
```
Nattalli, line 5: specify QuerySet type

services/movie.py
```diff
@@ -0,0 +1,43 @@
+from django.db.models import QuerySet
+from db.models import Genre, Actor, Movie
+
+
+def get_movies(
+        id_: int = None,
+        /,
```
Nattalli, line 7: ???

Nattalli (changes requested):

PR #1151, opened 2025-03-03
Solution

MarianaAB19:


db/migrations/0003_alter_movie_actors_alter_movie_genres_and_more.py
```diff
@@ -0,0 +1,34 @@
+# Generated by Django 4.0.2 on 2025-02-20 22:43
```
SerhiyTo, line 1: Try to stick to the development concept of ‘1 PR - 1 migration (per app)’. You don't need to do so many migrations in one iteration

SerhiyTo (changes requested):

SerhiyTo (approved):

PR #1169, opened 2025-03-19
Implement Movie, MovieSession, Cinemahall models and services

PrimeGlorious:


db/models.py
```diff
@@ -14,3 +14,39 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="movies")
+    genres = models.ManyToManyField(Genre, related_name="movies")
```
Boyok1337, line 23: Missed `on_delete`

db/models.py
```diff
@@ -14,3 +14,39 @@ class Actor(models.Model):
...
+
+    def __str__(self) -> str:
+        return self.title
+
+
+class CinemaHall(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
```
Boyok1337, line 32: Also can use `PositiveIntegerField`

db/models.py
```diff
@@ -14,3 +14,39 @@ class Actor(models.Model):
...
+        return f"{self.name}"
+
+
+class MovieSession(models.Model):
+    show_time = models.DateTimeField()
+    cinema_hall = models.ForeignKey(
+        CinemaHall,
+        on_delete=models.SET_NULL,
+        null=True
+    )
```
Boyok1337, line 48: Related name missing

db/models.py
```diff
@@ -14,3 +14,39 @@ class Actor(models.Model):
...
+    movie = models.ForeignKey(Movie, on_delete=models.SET_NULL, null=True)
```
Boyok1337, line 49: The same

services/movie.py
```diff
@@ -0,0 +1,47 @@
...
+            actors__id__in=actors_ids
+        )
+
+    if genres_ids:
+        return movies.filter(genres__id__in=genres_ids)
+
+    if actors_ids:
+        return movies.filter(actors__id__in=actors_ids)
+
+    return movies
```
Boyok1337, line 24: Simplify the logic here, you can generate 1 QS at once

services/movie_session.py
```diff
@@ -0,0 +1,59 @@
...
+
+def get_movies_sessions(
+        session_date: str = None
+) -> MovieSession:
+    if session_date:
+        return MovieSession.objects.filter(
+            show_time__date=session_date
+        )
+
+    return MovieSession.objects.all()
```
Boyok1337, line 28: Not the best approach, try to simplify like this
```
def get_movies_sessions(session_date: str = None) -> MovieSession:
    qs = MovieSession.objects.objects.all()
    if session_date:
        qs.filter(show_time__date=session_date)
    return qs
```

services/movie_session.py
```diff
@@ -0,0 +1,59 @@
...
+
+    session_to_update.save()
+
+    return session_to_update
+
+
+def delete_movie_session_by_id(
+        session_id: int,
+) -> None:
+    MovieSession.objects.filter(id=session_id).delete()
```
Boyok1337, line 59: If you delete by id = use only `.get`

Boyok1337 (changes requested):

Boyok1337 (approved):
GJ!

PR #1175, opened 2025-03-23
Solution

Nikita-pyth:


db/models.py
```diff
@@ -14,3 +14,38 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor, related_name="movies")
+    genres = models.ManyToManyField(Genre, related_name="movies")
```
AndyStarGitHub, line 23: Perhaps it would not be good to have the same related_names?

services/movie.py
```diff
@@ -0,0 +1,45 @@
...
+) -> QuerySet:
+    if not genres_ids and not actors_ids:
+        return Movie.objects.all()
+    if genres_ids and not actors_ids:
+        return Movie.objects.filter(genres__id__in=genres_ids)
+    if not genres_ids and actors_ids:
+        return Movie.objects.filter(actors__id__in=actors_ids)
+    return Movie.objects.filter(
+        genres__id__in=genres_ids,
+        actors__id__in=actors_ids
```
AndyStarGitHub, line 18: Perhaps it was possible to use fewer ifs. But it works - ok.

AndyStarGitHub (approved):
Good job!

PR #1181, opened 2025-03-23
Solution

OlenVe:


Kart0l (approved):
General code works, but in the Actor and CinemaHall classes, f-string is used where it is not necessary.
But anyway you did a good job!

PR #1196, opened 2025-03-27
task solution

yarms1:


services/movie.py
```diff
@@ -0,0 +1,48 @@
...
+        ).distinct()
+
+    return queryset
+
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
```
oliasymak24, line 31: You don’t need to `return` anything in create/update/delete methods

services/movie.py
```diff
@@ -0,0 +1,48 @@
...
+    elif genres_ids:
+        return queryset.filter(
+            genres__id__in=genres_ids
+        ).distinct()
+    elif actors_ids:
+        return queryset.filter(
+            actors__id__in=actors_ids
+        ).distinct()
+
+    return queryset
```
oliasymak24, line 24: distinct() you can use only once  in return

oliasymak24 (changes requested):

db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
...
+    genres = models.ManyToManyField(Genre, related_name="genres")
+
+    def __str__(self) -> str:
+        return self.title
+
+
+class CinemaHall(models.Model):
+    name = models.CharField(max_length=255)
+    rows = models.IntegerField()
+    seats_in_row = models.IntegerField()
```
Boyok1337, line 32: You can also use PositiveIntegerField

services/movie.py
Boyok1337: Still not fixed

Boyok1337 (changes requested):

services/movie.py
```diff
@@ -7,20 +7,13 @@ def get_movies(
...
-            actors__id__in=actors_ids
-        ).distinct()
-
+    if genres_ids or actors_ids:
+        filters = {}
+        if genres_ids:
+            filters["genres__id__in"] = genres_ids
+        if actors_ids:
+            filters["actors__id__in"] = actors_ids
+        return queryset.filter(**filters).distinct()
```
Boyok1337, line 16: You can write it easier, make it simple
Try smth like this
```
queryset = Movie.objects.all()
if genres_ids:
    queryset = queryset.filter(genres__id__in=genres_ids)
if actors_ids:
    queryset = queryset.filter(actors__id__in=actors_ids)
return queryset.distinct()
```

db/models.py
```diff
@@ -28,8 +28,8 @@ def __str__(self) -> str:
 
 class CinemaHall(models.Model):
     name = models.CharField(max_length=255)
-    rows = models.IntegerField()
-    seats_in_row = models.IntegerField()
+    rows = models.PositiveIntegerField()
+    seats_in_row = models.PositiveIntegerField()
```
Boyok1337, line 32: Good one!

Boyok1337 (approved):
Pay attention to `movie.py`

PR #1197, opened 2025-03-27
Anton Zhukov - Solution

Windings-Lab:


Windings-Lab:
@mateacademy-ai-mentor 
Tests are using IDs, so I can't using instances

PR #1223, opened 2025-04-14
Solution

olek007:


olek007:
@mateacademy-ai-mentor 
migrations are generated automaticly. The task is complete

olek007:
@mateacademy-ai-mentor 
Migrations was squashed into one as you suggest.

PR #1252, opened 2025-05-17
Solution

Michal-Naron:


Michal-Naron:
Now everything is correct because I adjust code based on your tips

PR #1308, opened 2025-07-12
implement movie, cinema hall, and movie session models and services

kravetst:


kravetb (approved):
Looks better than last one PR!

PR #1331, opened 2025-07-30
Solution

yagginov:


MakovM (approved):

yagginov:
@mateacademy-ai-mentor I don't affect the migration files, django does it for me and therefore I won't change it since it's not part of the task

PR #1335, opened 2025-07-30
Solution)))

Plovuezz:


services/movie.py
```diff
@@ -0,0 +1,40 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+
+
+def get_movies(
+        genres_ids: list = None,
+        actors_ids: list = None
+) -> QuerySet | Movie:
```
dimashevchukk, line 9: filter will always return QuerySet[Movie] in this case

services/movie.py
```diff
@@ -0,0 +1,40 @@
...
+
+    if genres_ids:
+        movie_query = movie_query.filter(genres__id__in=genres_ids)
+    if actors_ids:
+        movie_query = movie_query.filter(actors__id__in=actors_ids)
+
+    return movie_query
+
+
+def get_movie_by_id(movie_id: int) -> Movie | QuerySet:
```
dimashevchukk, line 21: get will always return only one Movie

services/movie.py
```diff
@@ -0,0 +1,40 @@
...
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        actors_ids: list = None,
+        genres_ids: list = None
+) -> None:
```
dimashevchukk, line 30: return Movie

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+from db.models import MovieSession, CinemaHall, Movie
+
+
+def create_movie_session(
+        movie_show_time: str,
+        movie_id: int,
+        cinema_hall_id: id
+) -> None:
```
dimashevchukk, line 11: return movie session

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
...
+    movie = Movie.objects.get(id=movie_id)
+
+    MovieSession.objects.create(
+        show_time=movie_show_time,
+        cinema_hall=cinema_hall,
+        movie=movie
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
```
dimashevchukk, line 23: QuerySet[MovieSession]

dimashevchukk (approved):

PR #1339, opened 2025-07-30
Solution

VadymKyrylko:


services/movie.py
```diff
@@ -0,0 +1,31 @@
+from db.models import Movie
+
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
+) -> list:
```
Sashoxxx, line 7: Not list but Queryset

services/movie_session.py
```diff
@@ -0,0 +1,53 @@
...
+    return MovieSession.objects.create(
+        movie_id=movie_id,
+        cinema_hall_id=cinema_hall_id,
+        show_time=movie_show_time
+    )
+
+
+def get_movies_sessions(
+        session_date: datetime = None
+) -> list:
```
Sashoxxx, line 19: Same

Sashoxxx (approved):
Good , but change returning annots mb.

PR #1346, opened 2025-07-30
Develop

bohdankolisnyk04:


DenisPrudius (approved):

services/movie.py
```diff
@@ -0,0 +1,29 @@
...
+
+def get_movies(genres_ids=None, actors_ids=None):
+    movies = Movie.objects.all()
+
+    if genres_ids and actors_ids:
+        movies = movies.filter(genres__id__in=genres_ids, actors__id__in=actors_ids).distinct()
+    elif genres_ids:
+        movies = movies.filter(genres__id__in=genres_ids).distinct()
+    elif actors_ids:
+        movies = movies.filter(actors__id__in=actors_ids).distinct()
```
MNDonut, line 12: you can remove the first condition and make 2 separate if statements because filters are appended and not run until you need it

services/movie_session.py
```diff
@@ -0,0 +1,38 @@
...
+
+def update_movie_session(session_id, show_time=None, movie_id=None, cinema_hall_id=None):
+    session = MovieSession.objects.get(id=session_id)
+
+    if show_time is not None:
+        session.show_time = show_time
+    if movie_id is not None:
+        session.movie_id = movie_id
+    if cinema_hall_id is not None:
+        session.cinema_hall_id = cinema_hall_id
```
MNDonut, line 31: what if user passes an empty string to any of these fields?

services/movie_session.py
```diff
@@ -0,0 +1,38 @@
...
+
+    session.save()
+    return session
+
+
+def delete_movie_session_by_id(session_id):
+    MovieSession.objects.get(id=session_id).delete()
```
MNDonut, line 38: what if movie session does not exist? Error will be raised by `.get` method

MNDonut (changes requested):
And tests failed

PR #1447, opened 2025-11-25
Solution

vakt159:


db/models.py
```diff
@@ -14,3 +15,37 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField(null=True)
+    actors = ManyToManyField(Actor)
+    genres = ManyToManyField(Genre)
```
Vorongor: Could be good to add a relative name for actors and genres filds

db/models.py
```diff
@@ -14,3 +15,37 @@ class Actor(models.Model):
...
+
+    def __str__(self) -> str:
+        return self.title
+
+
+class CinemaHall(models.Model):
+    name = models.CharField(max_length=255)
+    rows = models.IntegerField(null=True)
+    seats_in_row = models.IntegerField(null=True)
```
Vorongor: Is it ok, can you seat number be null? Better remove it or set default

services/movie.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+def get_movies(**kwargs) -> QuerySet:
+    genres_ids = kwargs.get("genres_ids")
+    actors_ids = kwargs.get("actors_ids")
+    queryset = Movie.objects.all()
+    if genres_ids:
+        queryset = queryset.filter(genres__id__in=genres_ids)
+    if actors_ids:
+        queryset = queryset.filter(actors__id__in=actors_ids)
```
Vorongor, line 11: Better option here would be using lookups: actors__id__in=actors_ids/genres_ids

services/movie.py
```diff
@@ -0,0 +1,27 @@
...
+    return queryset
+
+
+def get_movie_by_id(id_: int) -> Movie | None:
+    return Movie.objects.filter(id=id_).get()
```
Vorongor: Why don't you just use the get method?

services/movie_session.py
```diff
@@ -0,0 +1,35 @@
...
+                                movie_id=movie_id,
+                                cinema_hall_id=cinema_hall_id)
+
+
+def get_movies_sessions(*kwargs) -> QuerySet:
+    if not kwargs:
+        return MovieSession.objects.all()
+    else:
+        date_obj = datetime.strptime(kwargs[0], "%Y-%m-%d").date()
+        return MovieSession.objects.filter(show_time__date=date_obj)
```
Vorongor: You can get all elemnts by default and if you have a date just fiter that.

Vorongor (approved):
Good Job!
Overall, the structure looks correct, but highlighted several improvements:

Movie model: It’s recommended to add related_name to the actors and genres ManyToMany fields for clearer reverse relations.

CinemaHall model: rows and seats_in_row currently allow null=True. Reconsidering whether seat counts should ever be null, and possibly removing null=True or adding a default value.

Reverse relations (related_name="movie_session") are present, but naming them as plural ("movie_sessions") might better reflect Django conventions.

Filtering logic could use direct lookup syntax (e.g., actors__id__in) instead of accessing IDs in kwargs manually.

get_movie_by_id should use Movie.objects.get(id=id_) directly instead of filtering and chaining .get().

The date filtering logic can be simplified—first retrieving all sessions, then filtering if a date argument is provided.

PR #1448, opened 2025-11-25
Movie-moviesession-cinemahall

n-korobko:


db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
...
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    actors = models.ManyToManyField(Actor)
+    genres = models.ManyToManyField(Genre)
+
```
vakt159, line 24: Recommend you to add related_name in ManyToManyField

services/cinema_hall.py
```diff
@@ -0,0 +1,22 @@
...
+        hall_name: str,
+        hall_rows: int,
+        hall_seats_in_row: int
+) -> None:
+    cinema_hall = CinemaHall.objects.create(
+        name=hall_name,
+        rows=hall_rows,
+        seats_in_row=hall_seats_in_row
+    )
+    cinema_hall.save()
```
vakt159: I`m not sure, but you don`t need to call save(), because CinemaHall.objects.create should`ve already created new CInemaHall

services/movie.py
```diff
@@ -0,0 +1,47 @@
...
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
+) -> QuerySet:
+    movies = Movie.objects.all()
+    if genres_ids and actors_ids == "only":
+        return movies.filter(genres__id__in=genres_ids).distinct()
+
+    if actors_ids and genres_ids == "only":
+        return movies.filter(actors__id__in=actors_ids).distinct()
+
```
vakt159: Can`t understand logic of "only" in this case

vakt159 (approved):
Looks ok to me

db/models.py
petrykivd: Why u ignore this message? Add related name.

services/cinema_hall.py
```diff
@@ -0,0 +1,21 @@
+from django.db.models import QuerySet
+
+
+from db.models import CinemaHall
+
+
+def get_cinema_halls() -> QuerySet:
+    cinema_halls = CinemaHall.objects.all()
+    return cinema_halls
```
petrykivd, line 9: Simplify it and return in 1 line.

services/movie.py
```diff
@@ -0,0 +1,41 @@
...
+    if isinstance(actors_ids, (list, tuple)):
+        movies = movies.filter(actors__id__in=actors_ids)
+
+    return movies.distinct()
+
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    movies = Movie.objects.get(id=movie_id)
+    return movies
```
petrykivd, line 23: same here, can be in 1 line.  Useless variable movies

services/movie.py
```diff
@@ -0,0 +1,41 @@
...
+    movie = Movie.objects.create(
+        title=movie_title,
+        description=movie_description,
+    )
+    if genres_ids:
+        movie.genres.set(genres_ids)
+
+    if actors_ids:
+        movie.actors.set(actors_ids)
+    return movie
```
petrykivd, line 41: In function get_movies you check genres_ids does it instance of tuple or list. Why u don't check it here too?

petrykivd (changes requested):

PR #1449, opened 2025-11-26
add solution

VadymVynokur:


services/cinema_hall.py
```diff
@@ -0,0 +1,23 @@
...
+        hall_rows: int,
+        hall_seats_in_row: int
+) -> None:
+    cinema_hall = CinemaHall(
+        name=hall_name,
+        rows=hall_rows,
+        seats_in_row=hall_seats_in_row
+    )
+
+    cinema_hall.save()
```
stanislav-kishechko, line 23: Better in this case use create method

stanislav-kishechko (approved):

PR #1450, opened 2025-11-27
Solution

stanislav-kishechko:


db/models.py
```diff
@@ -14,3 +14,43 @@ class Actor(models.Model):
...
+        related_name="movie_sessions"
+    )
+    movie = models.ForeignKey(
+        Movie,
+        on_delete=models.CASCADE,
+        related_name="movie_sessions"
+    )
+
+    def __str__(self) -> str:
+        return f"{self.movie.title} {self.show_time}"
```
Guillelmina8, line 56: In requirements there was part: "There should be implemented string representation of the movie session,
that shows the movie name, **the date**, and the time of the movie session" and you implemented only title and time without date, not sure if this is really important but I'll leave it here

services/movie_session.py
```diff
@@ -0,0 +1,56 @@
...
+    )
+
+
+def get_movies_sessions(session_date: str = None) -> QuerySet:
+    queryset = MovieSession.objects.all()
+
+    if session_date is not None:
+        queryset = queryset.filter(show_time__date=session_date)
+
+    return queryset
```
Guillelmina8, line 26: Maybe you could also write `return queryset.filter(show_time__date=session_date)` without creating additional variable.

Guillelmina8 (approved):
Excellent work!

PR #1470, opened 2025-12-01
Solution

AIL8888:


services/cinema_hall.py
```diff
@@ -0,0 +1,19 @@
...
+
+def get_cinema_halls() -> Any:
+    return CinemaHall.objects.all()
+
+
+def create_cinema_hall(
+        hall_name: str,
+        hall_rows: int,
+        hall_seats_in_row: int
+) -> Any:
```
AngryAndrii: It's better to use more specific typing, because here we know that only Cinemahall object should returns

services/movie.py
```diff
@@ -0,0 +1,39 @@
...
+        queryset = queryset.filter(genres__id__in=genres_ids)
+
+    if actors_ids:
+        queryset = queryset.filter(actors__id__in=actors_ids)
+
+    return queryset.distinct()
+
+
+def get_movie_by_id(movie_id: int) -> Any:
+    return Movie.objects.get(id=movie_id)
```
AngryAndrii, line 22: here as well. We return only Movie object

AngryAndrii (approved):

PR #1601, opened 2026-04-18
created models and services

pawunder1992:


services/movie_session.py
```diff
@@ -0,0 +1,43 @@
...
+from db.models import MovieSession, CinemaHall, Movie
+
+
+def create_movie_session(movie_show_time: datetime.datetime,
+                         movie_id: int,
+                         cinema_hall_id: int) -> None:
+    cinema_hall = CinemaHall.objects.get(id=cinema_hall_id)
+    MovieSession.objects.create(show_time=movie_show_time,
+                                movie=Movie.objects.get(id=movie_id),
+                                cinema_hall=cinema_hall)
```
AnyoneClown: You don't need to fetch `cinema_hall` instance from db. Django ORM will automatically map FK to your `create` method, only add `cinema_hall_id`

services/movie_session.py
```diff
@@ -0,0 +1,43 @@
...
+
+def get_movies_sessions(session_date: datetime.datetime = None) -> QuerySet:
+    if session_date:
+        date = datetime.date(*list(map(int, session_date.split("-"))))
+        return MovieSession.objects.filter(show_time__date=date)
+    return MovieSession.objects.all()
+
+
+def get_movie_session_by_id(movie_session_id: int) -> MovieSession:
+    return MovieSession.objects.get(id=movie_session_id)
```
AnyoneClown: If movie doesn't exist - you will get an error. Use -> `from django.shortcuts import get_object_or_404`

services/movie_session.py
```diff
@@ -0,0 +1,43 @@
...
+                         movie_id: int = None,
+                         cinema_hall_id: int = None) -> None:
+    if show_time:
+        MovieSession.objects.filter(id=session_id).update(show_time=show_time)
+    if cinema_hall_id:
+        (MovieSession.objects.filter(id=session_id)
+         .update(cinema_hall=CinemaHall.objects.get(id=cinema_hall_id)))
+    if movie_id:
+        (MovieSession.objects.filter(id=session_id)
+         .update(movie=Movie.objects.get(id=movie_id)))
```
AnyoneClown: Refactor using only 1 query

AnyoneClown (changes requested):

AnyoneClown (approved):
LGTM

PR #1612, opened 2026-04-25
'Solution'

artradkevych:


db/models.py
```diff
@@ -1,4 +1,6 @@
 from django.db import models
+from django.db.models import ManyToManyField, ForeignKey
+from django.db.models import TextField, IntegerField, DateTimeField, CharField
```
pawunder1992, line 3: you dont need this, from django.db import models has it

pawunder1992 (approved):

PR #1700, opened 2026-07-07
'Solution'

dem1hh:


services/movie.py
```diff
@@ -0,0 +1,50 @@
...
+        ).distinct()
+        return queryset
+
+    if genres_ids:
+        queryset = queryset.filter(genres__id__in=genres_ids).distinct()
+        return queryset
+
+    if actors_ids:
+        queryset = queryset.filter(actors__id__in=actors_ids).distinct()
+        return queryset
```
hladiivolodymyr: You can simplify the logic here and avoid code duplication. Instead of checking every possible combination of genres_ids and actors_ids, apply each filter only if the corresponding parameter is provided, then return the queryset once at the end. This approach makes the code more concise, easier to read, and easier to maintain.

hladiivolodymyr (changes requested):

db/models.py
```diff
@@ -13,4 +13,44 @@ class Actor(models.Model):
...
+        related_name="movie_sessions"
+    )
+    movie = models.ForeignKey(
+        Movie,
+        on_delete=models.CASCADE,
+        related_name="movie_sessions"
+    )
+
+    def __str__(self) -> str:
+        return self.movie.title + " " + str(self.show_time)
```
vsmutok: use `f` string here

services/movie.py
```diff
@@ -0,0 +1,46 @@
...
+def get_movies(
+        genres_ids: list[int] = None,
+        actors_ids: list[int] = None
+) -> QuerySet:
+    queryset = Movie.objects.all()
+    if genres_ids and actors_ids:
+        queryset = queryset.filter(
+            genres__id__in=genres_ids,
+            actors__id__in=actors_ids
+        ).distinct()
```
vsmutok: This bit of code is unnecessary here.

vsmutok (changes requested):

db/models.py
```diff
@@ -13,4 +13,44 @@ class Actor(models.Model):
     last_name = models.CharField(max_length=255)
 
     def __str__(self) -> str:
-        return f"{self.first_name} {self.last_name}"
+        return self.first_name + " " + self.last_name
```
vsmutok: use `f` string here

services/movie.py
```diff
@@ -0,0 +1,44 @@
...
+    queryset = Movie.objects.all()
+
+    if genres_ids:
+        queryset = queryset.filter(genres__id__in=genres_ids)
+
+    if actors_ids:
+        queryset = queryset.filter(actors__id__in=actors_ids)
+
+    if genres_ids or actors_ids:
+        queryset = queryset.distinct()
```
vsmutok:
```suggestion
```

цей кусок коду тут зайвий

vsmutok (changes requested):

hladiivolodymyr (approved):

PR #1726, opened 2026-08-10
solution

perehinets:


services/movie_session.py
```diff
@@ -0,0 +1,44 @@
...
+                         cinema_hall_id: int
+                         ) -> MovieSession:
+    return MovieSession.objects.create(
+        show_time=movie_show_time,
+        cinema_hall_id=cinema_hall_id,
+        movie_id=movie_id
+    )
+
+
+def get_movies_sessions(session_date: str | None = None) -> MovieSession:
```
Sn1kls, line 17: Small note: this returns a `QuerySet`, not a single `MovieSession`, so the annotation is a bit off – `-> QuerySet[MovieSession]` (or just `-> QuerySet`) would be accurate. Not blocking

services/movie.py
```diff
@@ -0,0 +1,42 @@
+from django.db.models import QuerySet
+
+from db.models import Movie
+from typing import List
```
Sn1kls, line 4: Minor: since Python 3.9 you can use the built-in generics directly – `list[int]` instead of `typing.List[int]` – so this import isn't needed. Not blocking.

Sn1kls (approved):

PR #1729, opened 2026-08-12
Implement Movie, CinemaHall and MovieSession models

ibqqadir-art:
Add movie service functions
Add cinema hall service functions
Add movie session service functions
Add filtering, updating and deleting functionality


viktoria-rybenchuk (approved):
Good job!

PR #1732, opened 2026-08-17
movie-task-complicated

Dr-Kate:


db/models.py
```diff
@@ -14,3 +14,35 @@ class Actor(models.Model):
 
     def __str__(self) -> str:
         return f"{self.first_name} {self.last_name}"
```
vsmutok: The file uses f-string in Actor.__str__ (`return f"{self.first_name} {self.last_name}"`). The checklist discourages overusing f-strings for simple returns. Consider returning concatenated string or using formatting consistent with examples, e.g. `return self.first_name + " " + self.last_name` or `return "%s %s" % (self.first_name, self.last_name)` to follow the Good example guidance.

vsmutok (changes requested):

services/movie.py
```diff
@@ -0,0 +1,36 @@
...
+
+def get_movie_by_id(movie_id: int) -> Movie:
+    return Movie.objects.get(id=movie_id)
+
+
+def create_movie(
+        movie_title: str,
+        movie_description: str,
+        genres_ids: list = None,
+        actors_ids: list = None
```
Sn1kls: Small consistency note (non-blocking): the default here is `None`, so these should be `list | None = None`. You already do exactly this a few lines up in `get_movies (genres_ids: list | None = None)` – `create_movie` just got missed. Worth matching so the whole file is uniform. Everything else is great

Sn1kls (approved):
Approve. Nice work!)

